PR #2, opened 2022-04-22
Filetype checking & update Kafka connection to support new topics and messages structure

sixtedemaupeou:
- Only files detected as CSVs are downloaded to the datalake.
- Kafka connection is now compliant with https://www.notion.so/apigouv/Meetup-infra-des-donn-es-du-14-avril-2c5f4503ebd8417aae38928605a68af1


udata_datalake_service/background_tasks.py
```diff
@@ -39,33 +46,46 @@ def save_resource_to_minio(resource_file, key, resource):
...
 
 @celery.task
-def manage_resource(key, resource):
+def manage_resource(dataset_id: str, resource: dict):
     logging.info(
-        "Processing task for resource {} in dataset {}".format(resource["id"], key)
+        "Processing task for resource {} in dataset {}".format(resource["id"], dataset_id)
     )
     try:
         tmp_file = download_resource(resource["url"])
```
geoffreyaldebert, line 63: Can we use `resource["latest"]` instead of `url`. I think it is more secure.

udata_datalake_service/background_tasks.py
```diff
@@ -39,33 +46,46 @@ def save_resource_to_minio(resource_file, key, resource):
...
+                    f"Resource {resource['id']} in dataset {dataset_id} is not a CSV"
+                )
+
+        # Send a Kafka message for both CSV and non CSV resources
         logging.info(
-            "Sending kafka message for resource {} in dataset {}".format(resource["id"], key)
+            f"Sending kafka message for resource analysed {resource['id']} in dataset {dataset_id}"
         )
-        produce(key, resource)
+        message = {"mime": mime_type, "resource_url": resource["url"]}
```
geoffreyaldebert, line 87: same thing here. `resource["latest"]` seems more secure.

udata_datalake_service/background_tasks.py
```diff
@@ -39,33 +46,46 @@ def save_resource_to_minio(resource_file, key, resource):
...
+        produce("resource.analysed", resource["id"], message, meta={"dataset_id": dataset_id})
```
geoffreyaldebert, line 88: Can we name this topic `resource.analyzed` not `resource.analysed` (english)

udata_datalake_service/background_tasks.py
```diff
@@ -15,7 +17,7 @@
 
 BROKER_URL = os.environ.get("BROKER_URL", "redis://localhost:6380/0")
 MINIO_FOLDER = os.environ.get("MINIO_FOLDER", "folder")
-celery = Celery('tasks', broker=BROKER_URL)
+celery = Celery("tasks", broker=BROKER_URL)
 
 
 def download_resource(url):
```
geoffreyaldebert, line 23: Can we add in this function a test of the size of the resource.
We can set up a new env variable `MAX_FILESIZE_ALLOWED` (to 1Go for instance) and break out of the loop if we exceed the limit. 
We still need, even if the download does not succeed, a message in `resource.analyzed` a custom message that indicates that we could not download file because of size limitation.

udata_datalake_service/background_tasks.py
```diff
@@ -39,33 +46,46 @@ def save_resource_to_minio(resource_file, key, resource):
...
+        
+        # Check resource MIME type
+        mime_type = magic.from_file(tmp_file.name, mime=True)
+        if mime_type in ["text/plain", "text/csv"]:
+            # Save resource only if CSV
+            try:
+                # Raise ValueError if file is not a CSV
+                agate.Table.from_csv(tmp_file.name, sniff_limit=4096, row_limit=40)
+                save_resource_to_minio(tmp_file, dataset_id, resource)
+                storage_location = {"netloc": os.getenv("MINIO_URL"), "bucket": os.getenv("MINIO_BUCKET"), "key": MINIO_FOLDER + "/" + dataset_id + "/" + resource["id"]}
```
geoffreyaldebert, line 73: can we add in `resource.stored` message the `mime_type` of resource. It could be useful if we decide later to download other kind of files (for instance Excel files) but not process them for further analysis.

udata_datalake_service/consumer.py
```diff
@@ -8,14 +8,14 @@
...
-TOPICS = os.environ.get("TOPICS", ["dataset"])
+TOPICS = os.environ.get("TOPICS", ["resource.created", "resource.modified"])
 
 
 def create_kafka_consumer():
     logging.info("Creating Kafka Consumer")
     consumer = KafkaConsumer(
         bootstrap_servers=f"{KAFKA_HOST}:{KAFKA_PORT}",
-        group_id="datalake",
+        group_id=None,
```
geoffreyaldebert, line 18: Why do we set the group_id here to None ? I think group subscription to a specific group like `datalake` was a good idea.

geoffreyaldebert (approved):
👏  Very nice thanks. Approved if remarks are taken into account

geoffreyaldebert:
IMO next steps (not in this PR and not now) should be : 
- Manage Zip files and download them when csv are in it
- Manage Excel files

udata_datalake_service/background_tasks.py
```diff
@@ -39,33 +46,46 @@ def save_resource_to_minio(resource_file, key, resource):
...
-            )
+            f"Resource saved into minio at {get_resource_minio_url(key, resource)}"
         )
     except ClientError as e:
         logging.error(e)
 
 
 @celery.task
-def manage_resource(key, resource):
+def manage_resource(dataset_id: str, resource: dict):
```
quaxsze, line 58: 🤩

quaxsze (approved):

geoffreyaldebert:
A problem in this udata-datalake-service. As we set it up, process will try to download every resource. 
It could be a problem for specific platform like ODS which had quota for requesting data. That's why Decapode service has managed exception for specific domains to retrieve info : https://github.com/etalab/decapode/blob/b39ba823977d5c3ef2c53f62cb82c141d5849073/decapode/config.py#L4
We need to resolve that before production.

udata_datalake_service/background_tasks.py
maudetes: nevermind s is good

udata_datalake_service/consumer.py
sixtedemaupeou: Indeed it's a good idea, None was useful for testing purposes and I forgot to remove it. I will do so now.

geoffreyaldebert:
@quaxsze @sixtedemaupeou can we add in this PR information on message produce into kafka topic `resource.analysed` the actual `filesize` that we found downloading the resource ? If resource could not be downloaded, we can set the value of this property to `None` ?

PR #4, opened 2022-05-20
Add CircleCI

maudetes:
Add a CircleCI conf.
- Add a default MAX_FILESIZE_ALLOWED
- Add missing requirements
- Replace python-magic-bin by python-magic due to installation issues on Linux
- Fix some tests

TODO
- [ ] upgrade python docker image
---

Some things not treated:
- version based on CircleCI build in root `__init__.py` (useful if we want to push .devXXX on pypi)
- release process

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,10 +5,6 @@ jobs:
   build:
     docker:
       - image: udata/circleci:2-alpine
-        environment:
-          node.name: es01
-          cluster.name: es-docker-cluster
-          cluster.initial_master_nodes: es01
     environment:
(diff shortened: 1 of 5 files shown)


geoffreyaldebert (approved):
Approved but careful with this setup `python-magic` will only work on linux. `python-magic-bin` work on windows and mac. We should investigate later to know how to handle all envs with one spec of libs.

abulte:
I'm not sure we need to target anything else than Linux, a warning in README should be enough ;-)

sixtedemaupeou:
The contents of this PR are included in https://github.com/opendatateam/udata-datalake-service/pull/7

PR #25, opened 2022-08-29
Check by resource_id

sixtedemaupeou:
Instead of checking if the status has changed for the same URL we check for the same resource_id, so as to send a Kafka message whenever the resource availability has changed, regardless whether the URL was checked before for a different resource.

Fix https://github.com/etalab/udata-hydra/issues/19.

### DIFF
--- a/udata_hydra/crawl.py
+++ b/udata_hydra/crawl.py
@@ -49,6 +49,8 @@ async def insert_check(data: dict):
 async def update_check_and_catalog(check_data: dict) -> None:
     """Update the catalog and checks tables"""
     context.monitor().set_status("Updating checks and catalog...")
+    check_data['resource_id'] = str(check_data['resource_id'])
+
     pool = await context.pool()
     async with pool.acquire() as connection:
         q = f"""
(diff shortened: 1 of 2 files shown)


geoffreyaldebert (approved):

udata_hydra/crawl.py
```diff
@@ -227,19 +234,26 @@ async def check_url(row, session, sleep=0, method="get"):
...
+                "timeout": True,
+            }
+        )
+        return STATUS_TIMEOUT
     # TODO: debug AssertionError, should be caught in DB now
     # File "[...]aiohttp/connector.py", line 991, in _create_direct_connection
     # assert port is not None
     # UnicodeError: encoding with 'idna' codec failed (UnicodeError: label too long)
     # eg http://%20Localisation%20des%20acc%C3%A8s%20des%20offices%20de%20tourisme
-    except (
```
maudetes, line 235: Don't we want to keep these exhaustive exceptions, dealing with timeout separately? See also: https://github.com/etalab/udata-hydra/issues/17

maudetes (approved):
I have a small comment on exceptions, but except (:p) for this I'm good

udata_hydra/crawl.py
sixtedemaupeou: Indeed, I had not seen Alexandre's comment

PR #32, opened 2022-11-16
Compute checksum in checks

maudetes:
Close https://github.com/etalab/udata-hydra/issues/26

Compute checksum when downloading file.

Also add some enhancements:
* update resource url in `catalog` on resource modification
* store mime type, checksum, additional errors and filesize in `checks`

:warning: You should reinitialize your DBs with the new columns using `udata-hydra init-db --drop`


udata_hydra/datalake_service.py
```diff
@@ -131,3 +149,32 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
+            "mime_type": mime_type
+        }
+
+
+async def has_checksum_been_modified(resource_id, new_checksum):
+    q = """
+    SELECT
+        checksum
+    FROM checks
+    WHERE resource_id = $1 and checksum IS NOT NULL
```
maudetes, line 165: @geoffreyaldebert I've added `checksum IS NOT NULL` in the WHERE clause ;)

udata_hydra/config.py
```diff
@@ -26,4 +26,7 @@
 # -- Webhook integration config -- #
 ENABLE_WEBHOOK = True
 UDATA_URI = 'http://dev.local:7000/api/2/'
-UDATA_URI_API_KEY = ''
+UDATA_URI_API_KEY = None
+
+# Save resource to minio or not
+SAVE_TO_MINIO = False
```
maudetes, line 32: Fix https://github.com/etalab/udata-hydra/issues/33

I've added a simple config now to activate/deactivate saving resource to minio

udata_hydra/app.py
```diff
@@ -118,7 +118,8 @@ async def resource_updated(request):
         q = f"""SELECT * FROM catalog WHERE resource_id = '{resource_id}';"""
         res = await connection.fetch(q)
         if (len(res) != 0):
```
abulte, line 120:
```suggestion
        if len(res):
```

udata_hydra/crawl.py
```diff
@@ -218,6 +218,10 @@ async def check_url(row, session, sleep=0, method="get"):
                     "headers": convert_headers(resp.headers),
                     "timeout": False,
                     "response_time": end - start,
+                    "error": res['error'],
+                    "checksum": res['checksum'],
+                    "filesize": res['filesize'],
+                    "mime_type": res['mime_type']
```
abulte, line 224:
```suggestion
                    "error": res["error"],
                    "checksum": res["checksum"],
                    "filesize": res["filesize"],
                    "mime_type": res['mime_type']
```

udata_hydra/datalake_service.py
```diff
@@ -56,13 +58,23 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
+
     try:
         tmp_file = await download_resource(url, response)
 
         # Get file size
         filesize = os.path.getsize(tmp_file.name)
 
+        # Get checksum
+        with open(tmp_file.name, 'rb') as f:
+            sha1 = hashlib.sha1(f.read()).hexdigest()
```
abulte, line 76: Maybe nitpicking but we could compute sha1 in blocks instead of loading all the file at once (eg https://gist.github.com/techtonik/df09baeacbebc52d234b#file-sha1-py-L20)

udata_hydra/datalake_service.py
```diff
@@ -85,17 +97,16 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
-                        + "/"
-                        + resource_id,
-                    }
+                    storage_location = '/'.join([
+                        os.getenv("MINIO_URL"),
+                        os.getenv("MINIO_BUCKET"),
+                        MINIO_FOLDER,
+                        dataset_id,
+                        resource_id
+                    ])
```
abulte, line 109: ❤️

udata_hydra/datalake_service.py
```diff
@@ -115,14 +126,21 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
             'analysis:error': None,
             'analysis:filesize': filesize,
-            'analysis:mime': mime_type
+            'analysis:mime': mime_type,
         }
+        # Check if checksum has been modified
+        # TODO: improve file modification logic
+        checksum_modified = await has_checksum_been_modified(resource_id, sha1)
+        if checksum_modified:
+            document['analysis:checksum_last_modified'] = datetime.now().isoformat()
```
abulte, line 135: Can we move to a more [generic structure like here](https://github.com/etalab/udata-hydra/blob/fed48ab0b7e7180d3171afd27c0bb3fa00834be5/udata_hydra/app.py#L255)? If this breaks everything forget it, I'll do it later.

udata_hydra/datalake_service.py
```diff
@@ -115,14 +126,21 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
         document = {
             'analysis:error': None,
             'analysis:filesize': filesize,
```
abulte, line 128:
```suggestion
            "analysis:error": None,
            "analysis:filesize": filesize,
```

udata_hydra/datalake_service.py
```diff
@@ -115,14 +126,21 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
-            'analysis:mime': mime_type
+            'analysis:mime': mime_type,
```
abulte, line 129:
```suggestion
            "analysis:mime": mime_type,
```

udata_hydra/datalake_service.py
abulte, line 135:
```suggestion
            document["analysis:checksum_last_modified"] = datetime.now().isoformat()
```

abulte, line 135: Shouldn't we make sure we're using UTC? `utcnow()`

udata_hydra/datalake_service.py
```diff
@@ -115,14 +126,21 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
         await send(dataset_id=dataset_id,
                    resource_id=resource_id,
                    document=document)
     except IOError:
+        error = "File too large to download"
         document = {
-            'analysis:error': "File too large to download",
+            'analysis:error': error,
             'analysis:filesize': None,
             'analysis:mime': None,
```
abulte, line 145:
```suggestion
            "analysis:error": error,
            "analysis:filesize": None,
            "analysis:mime": None,
```

udata_hydra/datalake_service.py
```diff
@@ -132,3 +150,31 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
+    pool = await context.pool()
+    async with pool.acquire() as connection:
+        data = await connection.fetch(q, resource_id)
+        if data:
+            if data[0]['checksum'] != new_checksum:
+                return True
+            else:
+                return False
+        else:
+            # First check, thus this is a new checksum
```
abulte, line 179: Unsure about that 🤔 It doesn't mean the file has changed. We can iterate on `Last-Modified` header later if checksum is not conclusive.

udata_hydra/datalake_service.py
```diff
@@ -56,13 +58,23 @@ async def process_resource(url: str, dataset_id: str, resource_id: str, response
...
+    error = None
+
     try:
         tmp_file = await download_resource(url, response)
 
         # Get file size
         filesize = os.path.getsize(tmp_file.name)
 
+        # Get checksum
+        with open(tmp_file.name, 'rb') as f:
```
abulte, line 75:
```suggestion
        with open(tmp_file.name, "rb") as f:
```

udata_hydra/datalake_service.py
maudetes: Done :+1:

udata_hydra/datalake_service.py
maudetes: Indeed, it would be misleading to expose that the file has changed in this case.
I've modified the behaviour in https://github.com/etalab/udata-hydra/pull/32/commits/0fe46fc0352b7e38b15f4361cbac54cedde83696.

udata_hydra/datalake_service.py
maudetes: Done in https://github.com/etalab/udata-hydra/pull/32/commits/db281ed91270a70d1ddf3d88a903e0d8c1bf612f

quaxsze (approved):

abulte (approved):
I wonder if we should use MD5 instead of SHA1. Thus we could leverage Content-MD5 header https://www.rfc-editor.org/rfc/rfc1864.html. But I don't think it's widely used and we can iterate on that later.

maudetes:
> I wonder if we should use MD5 instead of SHA1. Thus we could leverage Content-MD5 header https://www.rfc-editor.org/rfc/rfc1864.html. But I don't think it's widely used and we can iterate on that later.

Indeed. We had chosen `sha1` because it is [the default one in udata](https://github.com/opendatateam/udata/blob/master/udata/core/dataset/models.py#L105) -> don't know if we would want to compare with values in udata. Let's merge this and iterate when we improve content modification detection!

PR #34, opened 2022-11-21
Use config module as centralized settings source

### LINKED ISSUE #30: Harmonize files which manage env variables
actually two files : 
- env file
- config.py file

The code uses in different use case one or other.

abulte:
Every config value is now accessed via `udata_hydra.config`, that may rely for some values on environment variables (defined in `.env` or not depending on your preferences or setup).

Fix #30


udata_hydra/utils/minio.py
```diff
@@ -1,50 +1,39 @@
...
-from dotenv import load_dotenv
 
-load_dotenv()
+from udata_hydra import config
+
 
 log = logging.getLogger("udata-hydra")
 
-MINIO_FOLDER = os.environ.get("MINIO_FOLDER", "folder")
+MINIO_FOLDER = config.MINIO_FOLDER
```
maudetes, line 13: Unused?

.env.sample
```diff
@@ -1,9 +1,7 @@
-UDATA_URI =
-UDATA_URI_API_KEY =
 MINIO_URL =
 MINIO_USER =
 MINIO_BUCKET =
 MINIO_PWD =
 MINIO_FOLDER =
-MAX_FILESIZE_ALLOWED =
```
maudetes, line 8: You should probably update the Readme accordingly

udata_hydra/config.py
```diff
@@ -1,3 +1,12 @@
+import os
+
+# -- general settings -- #
+LOG_LEVEL = "DEBUG"
```
maudetes, line 4: It would be nice to be able to override LOG_LEVEL with env I guess

udata_hydra/config.py
```diff
@@ -22,11 +31,18 @@
 SINCE = '1w'
 # seconds to wait for between batches
 SLEEP_BETWEEN_BATCHES = 60
```
maudetes, line 33: Could we make this env easily overridable? :pray:

udata_hydra/config.py
```diff
@@ -1,3 +1,12 @@
+import os
```
maudetes, line 1: Shouldn't we `load_dotenv()` at the beginning of the config file?

maudetes (commented):
Nice! Thank you for the clean :)

maudetes (approved):
Me parece bien!

PR #38, opened 2022-11-24
Sentry integration and logging improvements

### LINKED ISSUE #28: Ajouter l'intégration Sentry

abulte:
- All modules (cli, crawler, app) now use the same logging system
- Colored logs!
- Sentry integration, with a bit of rationalisation of error logging not to clutter the issues stream too much
- A small CLI command to check an URL, because obviously I've found and fixed a crawler crash while testing Sentry

Capture d’écran 2022-11-24 à 15 51 15

Capture d’écran 2022-11-24 à 16 19 24

Fix #28


pyproject.toml
```diff
@@ -21,6 +21,9 @@ python-dotenv = "^0.19.2"
 python-magic = "^0.4.25 "
 redis = "^4.1.4"
 requests = "^2.27.1"
+sentry-sdk = "^1.11.1"
+aiocontextvars = "^0.2.2"
```
maudetes, line 25: Do we need this for Sentry?

udata_hydra/cli.py
```diff
@@ -124,6 +122,23 @@ async def load_catalog(url=CATALOG_URL):
         os.unlink(fd.name)
 
 
+@cli
+async def check_url(url, method="get"):
+    """Quickly check an URL"""
+    log.info(f"Checking url {url}")
+    async with aiohttp.ClientSession(timeout=None) as session:
```
maudetes, line 129: Wouldn't we want to use the check_url core function? (We would need to factorize it)

maudetes (commented):
Nice! :ok_hand: 

Seeing your issues on Sentry, you should run `poetry run udata-hydra init-db --drop` if you haven't already.
Tables have been modified since https://github.com/etalab/udata-hydra/pull/32

maudetes (approved):

pyproject.toml
abulte: Yes

udata_hydra/cli.py
abulte: Don't worry, I've another helper coming for that ;-)

PR #40, opened 2022-11-30
Introduce a queuing system

abulte:
This introduces a queuing system, based on RQ. Only the most simple task is delayed for now, `send_check_data`. `process_resource` will be a good candidate for the next one. The sentry integration was quite difficult to configure properly but [works fine now](https://sentry.data.gouv.fr/organizations/sentry/issues/2234/?project=23&query=is%3Aunresolved)!

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,12 @@ There's a by-domain backoff mecanism. The crawler will wait when, for a given do
 
 If an URL matches one of the `EXCLUDED_PATTERNS`, it will never be checked.
 
+## Worker
+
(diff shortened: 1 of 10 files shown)


maudetes (commented):
Seems to work nicely!
Would we want to queue all calls to udata? :)

abulte:
@maudetes all calls to udata are queued in #41

abulte:
@maudetes maybe you want to review #41 directly, much of what is done here has been changed there.

Issue #29, opened 2022-11-16
Use an ORM?


abulte:
I guess we'll be writing more and more queries and an ORM could help making the code cleaner and avoiding errors. https://github.com/python-gino/gino is pretty nice and works with async.

**EDIT 2024/07/30 by @bolinocroustibat:**

Other favorite ORM options:
- [Piccolo ORM](https://piccolo-orm.com/) : async, many perks and integrated admin
- [Tortoise ORM](https://tortoise.github.io/) : supposedly the [fastest](https://github.com/tortoise/orm-benchmarks), async, quite popular (more stars on GitHub, and recommended by FastAPI creator - also I'm experienced on that one :) )

bolinocroustibat:
Since we already use SQLalchemy in other parts of the code, we should use this existing ORM

bolinocroustibat:
Not sure it's necessary anymore since https://github.com/datagouv/hydra/pull/55 and https://github.com/datagouv/hydra/pull/112 have been closed.

PR #42, opened 2022-12-05
Add basic migration system

abulte:
Pending #29, this allows us to write simple SQL migrations.

The [library](https://github.com/adriangb/asyncpg-trek) is unfortunately not published on pypi, but code is clean and simple. This should only be temporary anyway.

I made a small hack with the `_LATEST_REVISION` file in order to be able to upgrade to the latest revision (sic) without knowing its name.


abulte:
@maudetes asyncpg-trek is [now on pypi](https://github.com/etalab/udata-hydra/pull/45/commits/d95c531d3f815c33c2af3fe7bb398d4e4af0a80a)! And I'm an [official contributor](https://github.com/adriangb/asyncpg-trek/pull/1) ;-)

maudetes (approved):
Nice! I find the usage and the _LATEST_REVISION hack simple enough!

I added a very blocking comment, though :p

abulte:
@adriangb I have to admit I've never used a migration system with such a feature, so I'm not sure what use case it could apply to. Maybe ideas will come over time? We're not using it for complicated stuff right now. I don't known if it's linked but your library was by far the simplest (maintained) I've seen. I only had to resort to the `_LATEST_REVISION` file hack to make it work for us, since I did not want every dev and ops from the team to be aware of this latest revision.

PR #46, opened 2022-12-08
Refactor crawl process

abulte:
Various improvements in the crawling logic and code:
- uncluttered `update_check_and_catalog` and rename
- uncluttered `compute_check_has_changed` and logic fixes
- use `HEAD` requests by default for crawling, fallback on `GET`
- get rid of useless `initialization` column


udata_hydra/crawl.py
```diff
@@ -26,20 +26,22 @@
...
+
+
 async def compute_check_has_changed(check_data, last_check) -> bool:
-    # FIXME: this is flawed (eg ssl.SSLCertVerificationError won't have a status)
-    is_first_check = last_check["status"] is None
-    status_has_changed = (
-        "status" in check_data
-        and check_data["status"] != last_check["status"]
-    )
+    is_first_check = not last_check
```
maudetes, line 37: If we don't want to check for `last_check` in every condition, we could compute criterions only if last_check exists? Not sure it would make it cleaner though

udata_hydra/crawl.py
```diff
@@ -52,26 +54,28 @@ async def compute_check_has_changed(check_data, last_check) -> bool:
     has_changed = any(criterions.values())
     if has_changed:
         document = {
-            "check:status": check_data["status"] if status_has_changed else last_check["status"],
+            "check:valid-status": is_valid_status(check_data.get("status")),
+            "check:status": check_data.get("status"),
             "check:timeout": check_data["timeout"],
-            "check:check_date": datetime.utcnow().isoformat(),
+            "check:check-date": datetime.utcnow().isoformat(),
             "check:error": check_data.get("error"),
```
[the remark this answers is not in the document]
abulte: Good idea! Done (and tested because it wasn't 😭) https://github.com/etalab/udata-hydra/pull/46/commits/e81d5ad8d7eb62a4441ff093c4c1ef3050ae8bdc

udata_hydra/crawl.py
abulte: Not sure either, that would make an ugly if or mass definitions. I quite like the idea that each criterion is independent too.

tests/conftest.py
```diff
@@ -156,3 +159,8 @@ async def _fake_check(
         return data
 
     return _fake_check
+
+
+@pytest.fixture
+def udata_url():
```
maudetes, line 165: :pray:

maudetes (approved):
I think we're good!

PR #47, opened 2022-12-09
Send analysis results only if changed or first check

abulte:


udata_hydra/analysis.py
```diff
@@ -75,49 +74,76 @@ async def process_resource(check_id: int) -> None:
...
+                "checksum": dl_analysis.get("analysis:checksum"),
+                "analysis_error": dl_analysis.get("analysis:error"),
+                "filesize": dl_analysis.get("analysis:filesize"),
+                "mime_type": dl_analysis.get("analysis:mime-type"),
+            })
+
+    has_changed_over_time = await detect_has_changed_over_time(change_analysis, resource_id, check_id)
+
+    change_analysis.update(dl_analysis)
+    if has_changed_over_time or (is_first_check and change_analysis):
```
maudetes, line 110: Makes me think the wording of `change_analysis` may be a bit weird in the right part. Indeed, we're in the case of a first check but still care about "change" analysis (that also encapsulates dl analysis).

We could rename it to `last_modified_analysis` for example?

Additionally, we could have a third analysis object for readability?
`complete_analysis={**change_analysis, **dl_analysis}`

maudetes, line 110: Should we add some debug log here? Feels like it may be useful in the future to check why analysis has been sent or not to udata?

udata_hydra/analysis.py
```diff
@@ -75,49 +74,76 @@ async def process_resource(check_id: int) -> None:
...
+        SELECT detected_last_modified_at
+        FROM checks
+        WHERE resource_id = $1 AND detected_last_modified_at IS NOT NULL
+        ORDER BY created_at DESC
+        LIMIT 1
+        """
+        pool = await context.pool()
+        async with pool.acquire() as conn:
+            res = await conn.fetchrow(q, resource_id)
+            if res and res["detected_last_modified_at"] != last_modified:
```
maudetes, line 136: Shouldn't we cast last_modified to isoformat before comparing? (Not sure why it doesn't show in tests though)

maudetes (commented):
Nice! I'm adding some minor comments.But additionally, I'm scared we're having an issue on the second checks on a remote url that may have changed.

For example let's take a random binary url. On the first check, we don't detect a `detected_last_modified_at` to store on the checks table. On the second check, when we call `detect_has_changed_over_time`, it returns False because the query does not yield a result where `detected_last_modified_at IS NOT NULL`, even if checksum has been detected to change. But at the end we don't send this information.

udata_hydra/analysis.py
abulte: Nope, the ORM returns datetime objects for timestamp columns :-)

maudetes: My bad, I mean we're comparing datetime objects and isoformat string here?

abulte: Hum yes 😱 And it's working 🤷

abulte: Fixed!

udata_hydra/analysis.py
abulte: I feel it might be too verbose, cf https://github.com/etalab/udata-hydra/pull/48#discussion_r1046376640

abulte: Additionally or alternatively? I went for alternatively ;-)

abulte:
> For example let's take a random binary url. On the first check, we don't detect a detected_last_modified_at to store on the checks table. On the second check, when we call detect_has_changed_over_time, it returns False because the query does not yield a result where detected_last_modified_at IS NOT NULL, even if checksum has been detected to change. But at the end we don't send this information.

Right, I might have wanted to be too much generic by handling all `change_analysis` the same way :-/

abulte:
@maudetes fixed! fun fact, I already had the tests but wrote them wrong, I should have caught this.

maudetes (approved):
I think we're good! Thank you :pray:

PR #51, opened 2023-02-07
Add new types: json, date and datetime and notify udata of parsing

abulte:
Re https://github.com/etalab/csv-detective/pull/48, this adds support for new types when parsing CSV:
- json
- date
- datetime

I had to switch from `dateutil` to `dateparser` in order to parse localised dates. This led to a reflexion about aware vs naive date times, and I decided to go aware everywhere. This seems better for comparison purposes and leads to less head scratching while writing code (but not tests 😅). FYI: dates are stored in UTC by Postgres and are converted to/from aware date times in Python.


abulte:
I suddenly have a doubt about udata: can we send timezone aware iso dates in the API?

abulte:
@maudetes Long time no review! You get a bonus feature for your trouble: https://github.com/etalab/udata-hydra/pull/51/commits/a03d9cb49fa32c4bf657e1c151d778badd91ceff ;-) Now we ping udata when parsing is done.

udata_hydra/analysis/csv.py
```diff
@@ -43,18 +50,43 @@
...
+    resource_id = analysis["resource_id"]
+    db = await context.pool()
+    record = await db.fetchrow("SELECT dataset_id FROM catalog WHERE resource_id = $1", resource_id)
+    if record:
+        payload = {
+            "resource_id": resource_id,
+            "dataset_id": record["dataset_id"],
+            "document": {
+                "analysis:parsing:table": analysis["parsing_table"],
+                "analysis:parsing:error": analysis["parsing_error"],
```
maudetes, line 83: I'm adding this here because it occurred for me, but we can remove the pruning of `None` values in `send` now :)

maudetes (approved):
Nice! I've tested it with some sample data with dates and json as well as invalid csv and it seems to work nicely :)

PR #52, opened 2023-02-11
Remove csv_analysis, integrate into checks

abulte:
So I tried to add an API endpoint for analysis results and I realised the `csv_analysis` table was more of a hassle than anything — just like we discussed.

Thus `csv_analysis` is removed, interesting columns are integrated to `checks` and the `checks` API now handles csv analysis results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Current (in progress)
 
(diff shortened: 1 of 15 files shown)


maudetes (approved):
Approved! :clap: 
A big big thank you for this PR and all these PRs before! :pray:

Don't forget the changelog entry! ;)

PR #53, opened 2023-02-11
Add web_anon role for postgres

abulte:
Needed for https://github.com/etalab/udata-hydra-csvapi

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import asyncio
+import hashlib
 import os
 
 from datetime import datetime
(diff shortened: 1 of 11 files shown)


abulte:
I'm not sure this should be handled by hydra, Ansible would be nice too. But how to handle local dev?

bolinocroustibat:
Is this still needed?

It seems this is already at least partly implemented since then. I see `parsing_table`, `parsing_error` and `parsing_date` columns of `checks` DB table already exist in the current DB schema.

Should we close this issue then?

PR #55, opened 2023-02-17
Start db refactoring

abulte:
I've been thinking a bit about the need for pg schema support and the refactoring involved.

I think we would be better off by starting with refactoring all SQL queries to go through `db`, and add schema support from there.

A draft for a few queries in this PR. Alternative: #29.


abulte:
@maudetes @jordanguedj

jordanguedj:
FYI `asyncpg` allows to configure the search path: https://github.com/etalab/udata-hydra/commit/4896a8c64c55eee8e8e563f344cf03d6f08088a8

bolinocroustibat:
This is actually done in https://github.com/datagouv/hydra/pull/104 extremely similarly, so once this PR is closed we can closed that one as well.

bolinocroustibat:
This is actually done in https://github.com/datagouv/hydra/pull/112 extremely similarly, so closing this one.

PR #57, opened 2023-03-03
Fix/typo and schema

maudetes:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/392/?project=2&query=is%3Aunresolved&referrer=issue-stream
Also use schema set in DATABASE_SCHEMA in cli.

Not sure if we should keep the cli_wrapper connecting to main only though, since we now have 2 dbs

### DIFF
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -262,7 +262,7 @@ async def purge_csv_tables():
             await delete_table(table)
             await conn.execute(
                 "UPDATE checks SET parsing_table = NULL WHERE parsing_table = $1", table
-            )   
+            )
             count += 1
     if count:
         log.info(f"Deleted {count} table(s).")
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -195,7 +195,11 @@ async def drop_dbs(dbs=[]):
         if db == "main":
             conn = context["conn"]
         else:
-            conn = await asyncpg.connect(dsn=getattr(config, f"DATABASE_URL_{db.upper()}"))
+            conn = await asyncpg.connect(
+                dsn=getattr(config, f"DATABASE_URL_{db.upper()}"),
+                server_settings={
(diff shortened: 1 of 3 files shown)


udata_hydra/cli.py
```diff
@@ -261,7 +265,11 @@ async def purge_csv_tables():
 
 @wrap
 async def cli_wrapper():
-    context["conn"] = await asyncpg.connect(dsn=config.DATABASE_URL)
+    context["conn"] = await asyncpg.connect(
```
jordanguedj, line 268: Indeed, maybe we should remove the wrapper as we now have `DATABASE_URL` & `DATABASE_URL_CSV` as connection strings, it can be confusing to wrap for `DATABASE_URL` by default.

udata_hydra/cli.py
```diff
@@ -192,19 +206,10 @@ async def csv_sample(size=1000, download=False, max_size="100M"):
...
-                    "search_path": config.DATABASE_SCHEMA
-                })
-
-        tables = await conn.fetch("""
+        conn = await connection(db)
+        tables = await conn.fetch(f"""
             SELECT tablename FROM pg_catalog.pg_tables
-            WHERE schemaname != 'information_schema' AND
-            schemaname != 'pg_catalog';
+            WHERE schemaname = '{config.DATABASE_SCHEMA}';
```
maudetes, line 212: Instead of excluding some internal schema, we explicit our target schema

udata_hydra/cli.py
maudetes: Let me know if [this solution](https://github.com/etalab/udata-hydra/pull/57/commits/099db57864d13300f2e259fdd4dbf0ad827e1aab) suits you :)

jordanguedj: Looking nice! Thanks for this

jordanguedj (approved):
LGTM!

PR #68, opened 2023-05-16
Feat/support multi crawler

maudetes:
Today, we can't start two crawler processes, since they could select the same batch of urls to crawl in an iteration.

We're adding a `status` column in the catalog, set to `crawling` when the url has been selected, and reset to NULL after crawl.

### DIFF
--- a/udata_hydra/crawl.py
+++ b/udata_hydra/crawl.py
@@ -307,8 +307,29 @@ async def crawl_urls(to_parse):
 
 def get_excluded_clause():
     return " AND ".join(
-        [f"catalog.url NOT LIKE '{p}'" for p in config.EXCLUDED_PATTERNS]
+        [f"catalog.url NOT LIKE '{p}'" for p in config.EXCLUDED_PATTERNS] +
+        [
+            "deleted = False",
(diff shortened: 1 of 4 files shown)


udata_hydra/crawl.py
```diff
@@ -307,10 +314,40 @@ async def crawl_urls(to_parse):
 
 def get_excluded_clause():
     return " AND ".join(
-        [f"catalog.url NOT LIKE '{p}'" for p in config.EXCLUDED_PATTERNS]
+        [f"catalog.url NOT LIKE '{p}'" for p in config.EXCLUDED_PATTERNS] +
+        [
+            "catalog.deleted = False",
+            "(catalog.status <> 'crawling' OR catalog.status IS NULL)"
```
geoffreyaldebert, line 320: why `<>` and not `!=` ?

geoffreyaldebert (approved):
I think it will work 🚀 :)

udata_hydra/crawl.py
maudetes: It had to do with me struggling with NULL comparison, replaced with `!=` :)

PR #69, opened 2023-05-22
Add health route

maudetes:
Add health route, that assert Postgres connection and returns a 200.

The Postgres connection tests runs a `SELECT 1`, similar to what has been done in PostgREST [in their ready route](https://github.com/PostgREST/postgrest/blob/e572d1d1a2ea14cd4505be2fa0c2c08e46d32997/src/PostgREST/Workers.hs#L295).


jordanguedj (approved):
Thanks!

PR #73, opened 2023-06-05
Fix conflict on updating catalog with multiple entries for a resource

maudetes:
Fix https://github.com/etalab/udata-hydra/issues/74

Add a constraint on resource_id unique, instead of the tuple (dataset_id, resource_id, url).
It means that on any resource update, we update URL value accordingly.
Past URLs that have been checked are kept in the `checks` table, thus keeping a trace of previous checked URLs if needed.
We also update the `purge_csv_tables` logic, that was expecting outdated URLs to be marked as deleted in the catalog.
But we can actually have URLs that disappear from the catalog, updated to the latest URL used for this resource.

Add unit test to describe buggy behavior
* There is one test on the load_catalog logic that creates multiple entries for a single resource
* And another test to raise a 500 on api udpate in case of multiple resource entries
* Add a test case on purge_csv_tables with url not in catalog anymore

### DIFF
--- a/udata_hydra/app.py
+++ b/udata_hydra/app.py
@@ -94,7 +94,10 @@ async def resource_created(request):
         q = f"""
                 INSERT INTO catalog (dataset_id, resource_id, url, deleted, priority)
                 VALUES ('{dataset_id}', '{resource_id}', '{resource["url"]}', FALSE, TRUE)
-                ON CONFLICT (dataset_id, resource_id, url) DO UPDATE SET priority = TRUE;"""
+                ON CONFLICT (resource_id) DO UPDATE SET
+                    priority = TRUE,
+                    url = '{resource["url"]}',
+                    dataset_id = '{dataset_id}';"""
         await connection.execute(q)
 
     return web.json_response({"message": "created"})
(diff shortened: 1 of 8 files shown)


udata_hydra/app.py
```diff
@@ -94,7 +94,10 @@ async def resource_created(request):
         q = f"""
                 INSERT INTO catalog (dataset_id, resource_id, url, deleted, priority)
                 VALUES ('{dataset_id}', '{resource_id}', '{resource["url"]}', FALSE, TRUE)
-                ON CONFLICT (dataset_id, resource_id, url) DO UPDATE SET priority = TRUE;"""
+                ON CONFLICT (resource_id) DO UPDATE SET
+                    priority = TRUE,
+                    url = '{resource["url"]}',
+                    dataset_id = '{dataset_id}';"""
```
geoffreyaldebert, line 100: I think it is not useful for now but maybe one day...

udata_hydra/cli.py
```diff
@@ -247,30 +250,34 @@ async def purge_checks(limit=2):
...
-            SELECT md5(url) FROM catalog WHERE deleted = TRUE
-        )
+    SELECT checks.parsing_table
+    FROM checks
+    LEFT JOIN (
+        select url, MAX(id) as id, BOOL_AND(deleted) as deleted
+        FROM catalog
+        GROUP BY url) c
+    ON checks.parsing_table = md5(c.url)
+    WHERE checks.parsing_table IS NOT NULL AND (c.id IS NULL OR c.deleted = TRUE);
```
geoffreyaldebert, line 267: will there be a problem here if two resources (two ids) share the same url ?

geoffreyaldebert: nevermind it won't matter with this query

geoffreyaldebert (approved):

PR #78, opened 2023-07-27
Compare content type / length to check if changed

maudetes:
In order to send to udata in compute_check_has_changed.

Indeed, `check:headers:content-length` is not updated if has changed since it wasn't taken into account in compute_check_has_changed.
Same for `check:headers:content-type`.


geoffreyaldebert (approved):
seems nice 🙏

PR #84, opened 2023-11-17
Enable csv.gz analysis

Pierlou:
This enables the analysis and database loading of csv.gz files (compressed csv)

### DIFF
--- a/udata_hydra/utils/csv.py
+++ b/udata_hydra/utils/csv.py
@@ -2,10 +2,13 @@ import json
 
 
 async def detect_csv_from_headers(check) -> bool:
-    """Determine if content-type header looks like a csv's one"""
-    headers = json.loads(check["headers"] or {})
+    """
+    Determine if content-type header looks like a csv's one
+    or if it's binary for potential csv.gz
+    """
+    headers = json.loads(check["headers"] or "{}")
     return any(
         headers.get("content-type", "").lower().startswith(ct) for ct in [
             "application/csv", "text/plain", "text/csv"
         ]
-    )
+    ), headers.get("content-type", "").lower().startswith("application/octet-stream")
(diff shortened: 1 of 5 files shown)


udata_hydra/utils/csv.py
```diff
@@ -2,10 +2,13 @@
...
+    Determine if content-type header looks like a csv's one
+    or if it's binary for potential csv.gz
+    """
+    headers = json.loads(check["headers"] or "{}")
     return any(
         headers.get("content-type", "").lower().startswith(ct) for ct in [
             "application/csv", "text/plain", "text/csv"
         ]
-    )
+    ), headers.get("content-type", "").lower().startswith("application/octet-stream")
```
maudetes, line 14: Hmm this may end up downloading a lot of files. Can't we base our assumption on the file name in the URL instead? Not sure if any other hint could help?
cc @geoffreyaldebert

maudetes (commented):
Nice! It does seem to work with quite few added lines!
Could you please add some tests on a `csv.gz` file?

udata_hydra/utils/csv.py
Pierlou: Fixed in https://github.com/etalab/udata-hydra/pull/84/commits/3fa8dbafe557f0f971efb950d6234bf505085d25 : we have an additionnal check on the URL

udata_hydra/utils/csv.py
```diff
@@ -2,10 +2,17 @@
 
 
 async def detect_csv_from_headers(check) -> bool:
```
maudetes, line 4: Can't we compute the `is_csvgz` in this method? And return `is_csv or is_csvgz` independently?
Same for `download_resource` returning an additional `is_csv`, not sure it is necessary.
I think we could keep the complexity inside the functions only if possible

maudetes (approved):
Perfect, thank you! :pray:

PR #85, opened 2023-11-24
feat: create a list of exceptions to analyse despite larger size

Pierlou:


udata_hydra/utils/file.py
```diff
@@ -22,23 +22,23 @@ def compute_checksum_from_file(filename):
     return sha1sum.hexdigest()
 
 
-async def download_resource(url: str, headers: dict) -> BinaryIO:
+async def download_resource(url: str, headers: dict, exception_file: bool = False) -> BinaryIO:
```
maudetes, line 25: I would rename the new param `ignore_size` maybe?

maudetes (commented):
Could you please add a test? :pray:

tests/test_csv_analysis.py
```diff
@@ -60,6 +61,31 @@ async def test_analyse_csv_big_file(setup_catalog, rmock, db, fake_check, produc
...
 
+@pytest.mark.slow
+async def test_exception_analysis(setup_catalog, rmock, db, fake_check, produce_mock):
+    """
+    Tests that exception resources (files that are too large to be normally processed) are indeed processed.
+    """
+    config.override(MAX_FILESIZE_ALLOWED=5000)
+    check = await fake_check(resource_id=config.LARGE_RESOURCES_EXCEPTIONS[0])
+    await db.execute(f"UPDATE catalog SET resource_id = '{config.LARGE_RESOURCES_EXCEPTIONS[0]}' WHERE id=1")
+    await db.execute("UPDATE catalog SET last_check = 1 WHERE id=1")
```
maudetes, line 72: If you UPDATE catalog with resource id before fake_check, you don't need to modify last_check, do you?

maudetes (approved):
Nice! The resulting code seems readable!

You can maybe add an entry in the changelog :pray:

tests/test_csv_analysis.py
Pierlou: Indeed, I'll change this!

PR #86, opened 2023-12-18
Add worker default timeout config

maudetes:
We may need to use a different default timeout for csv-to-db jobs, due to poor performances in csv-detective : https://github.com/etalab/csv-detective/issues/68.
Hopefully, we should be able to keep the default timeout in the long run.


Pierlou (approved):
🙏

PR #87, opened 2024-01-11
Return None value early in csv-to-db

maudetes:
Returns early with None when converting empty values to int, float, etc.

Currently, `smart_cast` first fails when trying `int("")`, then we try to `str2float("" default=None)`, and finally return None.
It ends up logging lots of errors:
```
WARNING Could not convert "" to int, defaulting to null
```
We also gain significant time (from ~30 to ~20 seconds on big file test): https://github.com/etalab/udata-hydra/blob/1a1361e68055f2a12df640a35b4c9e777298b38b/tests/test_csv_analysis.py#L41

[Related error](https://errors.data.gouv.fr/organizations/sentry/issues/4169/?project=21&query=&referrer=issue-stream&statsPeriod=14d#breadcrumbs) (that failed due to task timeout).

### DIFF
--- a/udata_hydra/analysis/csv.py
+++ b/udata_hydra/analysis/csv.py
@@ -115,6 +115,8 @@ def generate_dialect(inspection: dict) -> stdcsv.Dialect:
 
 def smart_cast(_type, value, failsafe=False) -> Any:
     try:
+        if value is None or value == "":
+            return None
         if _type == "bool":
             return str2bool(value)
         return PYTHON_TYPE_TO_PY[_type](value)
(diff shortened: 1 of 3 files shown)


udata_hydra/analysis/csv.py
```diff
@@ -115,6 +115,8 @@ class CustomDialect(stdcsv.unix_dialect):
 
 def smart_cast(_type, value, failsafe=False) -> Any:
     try:
+        if value is None or value == "":
```
Pierlou, line 118: Why not `if not value`? Is there a value that doesn't match this criterion?

Pierlou (approved):

udata_hydra/analysis/csv.py
maudetes: I was not sure of `value` possible types, but I didn't want to return None if the value was not empty but evaluated to False (False, [], etc.).

Pierlou: Makes sense 👍

PR #91, opened 2024-02-02
feat: draft ping to udata after apification

Pierlou:


Pierlou:
Restoring https://github.com/etalab/udata-hydra/pull/62 now that performances have increased

udata_hydra/analysis/csv.py
```diff
@@ -97,6 +121,8 @@ async def analyse_csv(check_id: int = None, url: str = None, file_path: str = No
                 "parsing_finished_at": datetime.now(pytz.UTC),
             })
         await csv_to_db_index(table_name, csv_inspection, check)
+        if check_id:
```
maudetes, line 124: We could move these lines in the `finally` section, so that we also ping parsing error info

udata_hydra/analysis/csv.py
```diff
@@ -65,6 +67,28 @@
...
+    check = await get_check(check_id)
+    resource_id = check["resource_id"]
+    db = await context.pool()
+    record = await db.fetchrow("SELECT dataset_id FROM catalog WHERE resource_id = $1", resource_id)
+    if record:
+        payload = {
+            "resource_id": resource_id,
+            "dataset_id": record["dataset_id"],
+            "document": {
+                "analysis:parsing:table": check["parsing_table"],
```
maudetes, line 81: @geoffreyaldebert should we keep `parsing:table` since we're using resource_id as the pivot value for tabular api?

maudetes: Upon discussion, we can remove `parsing:table` :)

maudetes (approved):
Looks good!
Could you add an entry in the changelog! :pray:

PR #92, opened 2024-03-14
Handle other file formats

Pierlou:
xls(x) can be inspected and uploaded to db
/!\ wait until new csv-detective version is released, currently working with a dev version


tests/test_csv_analysis.py
```diff
@@ -62,12 +62,13 @@ async def test_analyse_csv_big_file(setup_catalog, rmock, db, fake_check, produc
     assert profile["total_lines"] == expected_count
 
 
-@pytest.mark.slow
+# @pytest.mark.slow
```
maudetes, line 65:
```suggestion
```

tests/test_other_formats_analysis.py
```diff
@@ -0,0 +1,61 @@
+import hashlib
```
maudetes, line 1: The name of the file though :stuck_out_tongue:

At some point, we may put all in `test_tabular_analysis.py` and dispatch these in `CsvAnalysisTest` &co classes?

tests/test_other_formats_analysis.py
```diff
@@ -0,0 +1,61 @@
...
+import json
+import pytest
+
+from udata_hydra.analysis.csv import analyse_csv
+
+pytestmark = pytest.mark.asyncio
+
+
+async def test_csvgz_analysis(setup_catalog, rmock, db, fake_check, produce_mock):
```
maudetes, line 10: You could use `pytest.mark.parametrize` on filename and expected count between the tests?
[Example use](https://github.com/opendatateam/udata-metrics/blob/edfd98e152aa7e28b381520cdd936bfccf6563bd/tests/test_metrics.py#L15).

udata_hydra/utils/csv.py
```diff
@@ -1,21 +1,43 @@
...
         ]
-    ]) and "csv.gz" in check.get("url", []))
+    ]) and "csv.gz" in check.get("url", ""):
+        return True, "csvgz"
+
+    if any([
+        headers.get("content-type", "").lower().startswith(ct) for ct in [
+            "application/vnd.ms-excel",
+        ]
+    ]):
```
maudetes, line 31: You don't need to create a list, a generator is better here :)
```suggestion
    if any(
        headers.get("content-type", "").lower().startswith(ct) for ct in [
            "application/vnd.ms-excel",
        ]
    ):
```

udata_hydra/utils/file.py
```diff
@@ -31,23 +31,29 @@ def read_csv_gz(file_path):
     return temp_file
 
 
-async def download_resource(url: str, headers: dict, ignore_size: bool = False) -> BinaryIO:
+async def download_resource(
+    url: str,
+    headers: dict,
+    file_format: str,
+    ignore_size: bool = False
```
maudetes, line 38: `file_format` with ignore_size logic could be computed to use a single `max_size_allowed` param, that could be None?
It would make download resource slightly more focused?

udata_hydra/utils/reader.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            self.file = open(self.file_path, encoding=self.inspection["encoding"])
+            self.reader = stdcsv.reader(
+                self._skip_rows(),
+                dialect=generate_dialect(self.inspection)
+            )
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        if self.file is not None and hasattr(self.file, "close"):
```
maudetes, line 49: When there is no `close` method defined, it should close on its own, right?

udata_hydra/utils/reader.py
```diff
@@ -0,0 +1,67 @@
...
+
+def generate_dialect(inspection: dict) -> stdcsv.Dialect:
+    class CustomDialect(stdcsv.unix_dialect):
+        # TODO: it would be nice to have more info from csvdetective to feed the dialect
+        # in the meantime we might want to sniff the file a bit
+        delimiter = inspection["separator"]
+    return CustomDialect()
+
+
+class Reader:
```
maudetes, line 15: I think the Reader class very easy to read, thank you! :clap:

udata_hydra/analysis/csv.py
```diff
@@ -103,7 +105,13 @@ async def analyse_csv(check_id: int = None, url: str = None, file_path: str = No
     exception_file = str(check.get("resource_id", "")) in exceptions
 
     headers = json.loads(check.get("headers") or "{}")
-    tmp_file = open(file_path, "rb") if file_path else await download_resource(url, headers, exception_file)
+    tmp_file = open(file_path, "rb") if file_path else await download_resource(
+        url=url,
+        headers=headers,
+        # do we want to detect the format here?
```
maudetes, line 111: I don't think so. But, what we actually want here is a max size we want to download here. Using CSV's one is good with me for debug.

maudetes (commented):
Nice! I think it doesn't bloat the code any more than the current situation, well done!

I haven't tested it yet but would like to take the time to do so.

udata_hydra/utils/reader.py
Pierlou: I guess so

maudetes (approved):
I think we're all clear! Thank you again for taking time to add more tabular format support :rocket:

PR #27, opened 2022-10-25
Removed kafka integration, replaced by webhooks

quaxsze:
Fix https://github.com/etalab/data.gouv.fr/issues/982
Fix https://github.com/etalab/data.gouv.fr/issues/984
Fix #14

PR #101, opened 2024-07-16
Python 3.11 and cleaning

bolinocroustibat:
Some optional preparation before ORM refactoring:

- Use Python 3.11 for performance improvements and future compatibility
- Remove `toml` library and replaces by Python native `tomllib` library
- Remove `pandas` as project dependency, as it's only a subdependency of `csv-detective` dependency
- Remove useless typing imports
- Add type hints
- Remove deprecated version field in docker compose files
- Update `.gitignore`


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "udata-hydra"
 version = "2.0.0.dev"
 description = "Async crawler and parsing service for data.gouv.fr"
 authors = [{ name = "Opendata Team", email = "opendatateam@data.gouv.fr" }]
-requires-python = ">=3.9,<3.13"
+requires-python = ">=3.11,<3.13"
```
maudetes, line 6: Is it only due to tomllib that it isn't compatible with 3.9? Do we want to add a if-case to keep 3.9 support (making migration easier maybe).

udata_hydra/context.py
```diff
@@ -14,7 +14,7 @@
 }
 
 
-def monitor():
+def monitor() -> MagicMock:
```
maudetes, line 17: I think Monitor [died here](https://github.com/datagouv/hydra/pull/27) in silence. I had not realize it is only a Mock now ^^

maudetes (approved):
Looks good to me! :clap: Thank you for the cleaning phase!

pyproject.toml
bolinocroustibat: Yes, just because of that. So that we could remove a dependency and use `tomlib` in the standard library.
As you wish, I can't judge yet if we want to favor retrocompatibility towards 3.9, or lighten the dependencies :) Let me know and I'll commit a revert on this precise thing if necessary

maudetes: Since the actual version is already working with 3.11, we can drop 3.9 without making it too hard to upgrade step by step. :ok_hand:

bolinocroustibat: In that case I'll create another cleaning PR without the big move to Python 3.11, and rebase that one later one, while we keep that one on pause for now.

PR #102, opened 2024-07-22
clean: basic cleaning

bolinocroustibat:
Replaces https://github.com/datagouv/hydra/pull/101, some cleaning without fully moving to Python 3.11 :
- Use Python 3.11 in CI, in order to prepare to use 3.11 on production as well. Without changing the code so that it stays compatible with 3.9.
- Remove pandas as project dependency, as it's only a subdependency of csv-detective dependency
- Add type hints
- Fix some wrong type hints
- Remove deprecated version field in docker compose files
- Update .gitignore


.gitignore
```diff
@@ -1,4 +1,4 @@
-config.toml
+# VSCode
```
magopian, line 1: It feels a bit weird having user specific stuff in the project's `.gitignore` (users can have their own config in their `~/.gitignore`).

I know you only added the comment, not the line itself, so i'm not asking you to remove it, just a side note ;)

udata_hydra/analysis/csv.py
```diff
@@ -99,9 +99,9 @@ async def notify_udata(check_id):
 
 
 async def analyse_csv(
-    check_id: int = None,
-    url: str = None,
-    file_path: str = None,
+    check_id: int | None = None,
+    url: Union[str, None] = None,
```
magopian, line 103: Why use the `Union[str, None]` version, and not its shorthand (which seems to [be recommended](https://docs.python.org/3/library/typing.html#typing.Union)) `str | None`?

I'm new to python typing annotations, so I may be missing something obvious here 😬

udata_hydra/migrations/__init__.py
```diff
@@ -11,7 +11,7 @@
 class Migrator:
     # NB: we can't use async __init__, hence the workaround
     @classmethod
-    async def create(cls, db_name, skip_errors=False):
+    async def create(cls, db_name, skip_errors=False) -> "Migrator":
```
magopian, line 14: I understand you're using a string here as a forward reference to the type, but reading https://peps.python.org/pep-0649/ it feels that we can now simply use `Migrator` here? Maybe I'm mistaken, and it's not implemented yet?

magopian (approved):
Nice cleaning up and typing! Thanks 🙏

.gitignore
bolinocroustibat: You mean stuff related to a specific IDE in a `.gitignore` file? You think `.gitignore` should be IDE-agnostic? I think it can host stuff related to several possible dev environments, but open to discuss!

udata_hydra/analysis/csv.py
bolinocroustibat: You're totally right, but we decided with @maudetes that we would keep hydra as Python 3.9 compatible first (and "|" is not yet comaptible with Python 3.9), and we will later move to a nicer Python>=3.10 syntax!

udata_hydra/migrations/__init__.py
bolinocroustibat: My type checker doesn't accept that, from what I understand https://peps.python.org/pep-0649/ has been accepted but will only be implemented in Python 3.14?

udata_hydra/analysis/csv.py
bolinocroustibat: FYI there is another branch to fully switch to Python 3.11 including nicer typing not compatible with 3.9, but we decided to postpone it: https://github.com/datagouv/hydra/tree/py3.11

magopian: but, line 102 uses the shorthand `int | None` version, right?

.gitignore
magopian: yeah, it's kinda weird having IDE specific stuff in the project's `.gitignore` file, because we might then want to add all sorts of other IDE quirks and stuff... I'm not against it, I just find it odd

udata_hydra/analysis/csv.py
bolinocroustibat: Oh indeed, thanks for spotting, fixing that now

bolinocroustibat: Fixed

PR #107, opened 2024-07-29
fix: fix wrong types returns

bolinocroustibat:
Small fix regarding wrong types returns.

_Note:
In order to comply with the Python 3.9 retro-compatibility, we use only Python 3.9-compatible typings for now. The switch to cleaner >=3.10 typings are all gathered in the following draft PR once we switched to Python 3.11: https://github.com/datagouv/hydra/pull/108_

### DIFF
--- a/udata_hydra/analysis/helpers.py
+++ b/udata_hydra/analysis/helpers.py
@@ -11,7 +11,7 @@ def to_json(value: str) -> str:
     return value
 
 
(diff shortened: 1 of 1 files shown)


magopian (approved):
👍 
🚢

ThibaudDauce (approved):
:+1:

PR #110, opened 2024-07-30
docs: update changelog

bolinocroustibat:
Update changelog to take into account the recent changes that have been previously merged without changelog entries.


maudetes (approved):
:pray:

PR #117, opened 2024-07-31
refactor: refactor routes

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/115

Refactor routes URL, routes handle names and routes files.

Existing hydra routes were not very RESTful and not consistent in terms of URLs and methods. This PR refactors them, while keeping the old routes working to keep retrocompatibility.
- Tests have been adapted to test both new routes and legacy routes
- #TODO flags has been added to flag legacy routes that can be removed once they are not used anymore from any external service (udata)

One should make a new PR to clean legacy routes from hydra once udata will be modified to use the new routes.


udata_hydra/routes/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+routes.extend(
+    [
+        # Routes for checks
+        web.get("/api/checks/latest/", get_latest_check),
+        web.get("/api/checks/all/", get_all_checks),
+        # Routes for resources
+        web.post("/api/resources/", create_resource),
+        web.put("/api/resources/", update_resource),
+        web.delete("/api/resources/", delete_resource),
+        web.post("/api/resource/created/", create_resource),  # TODO: legacy, to remove
```
maudetes, line 19: I think you should probably update the Readme to document the new routes and the old ones deprecation!

tests/test_api.py
```diff
@@ -152,19 +116,30 @@ async def test_api_resource_updated(client):
             "last_modified": datetime.now().isoformat(),
         },
     }
-    resp = await client.post("/api/resource/updated/", json=payload)
+    client_http_method: Callable = getattr(
+        client, route["method"]
+    )  # TODO: can be removed once we don't use legacy route anymore
```
magopian, line 121: Nit: redundant `# TODO` as you already have it on line 99 before the test itself?

udata_hydra/routes/__init__.py
```diff
@@ -0,0 +1,27 @@
+from aiohttp import web
+
+from udata_hydra.routes.checks import get_all_checks, get_latest_check
+from udata_hydra.routes.resources import create_resource, delete_resource, update_resource
+from udata_hydra.routes.status import get_crawler_status, get_stats, get_worker_status
+
+# routes = web.RouteTableDef()
```
magopian, line 7: Is that a leftover comment that should be removed?

udata_hydra/routes/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+routes: list = []
+
+routes.extend(
+    [
```
magopian, line 11: Is there a reason to assign the routes on a new line, instead of 
```python
routes: list = [...]
```
?

magopian (approved):
Awesome job, looks way better like that, and it's a nice touch to have kept the legacy routes (and tests!).

I'm wondering if instead of the `# TODO`, we should use [the deprecation system](https://peps.python.org/pep-0702/) (and maybe some help from [the deprecation lib](https://deprecation.readthedocs.io/en/latest/))?

I've never used that so I can't speak from experience if that is a good idea, it just got me curious ;) (maybe in another PR?)

maudetes (approved):
Well done with the refactoring!

Don't forget to add a changelog entry!

bolinocroustibat:
> Awesome job, looks way better like that, and it's a nice touch to have kept the legacy routes (and tests!).
> 
> I'm wondering if instead of the `# TODO`, we should use [the deprecation system](https://peps.python.org/pep-0702/) (and maybe some help from [the deprecation lib](https://deprecation.readthedocs.io/en/latest/))?
> 
> I've never used that so I can't speak from experience if that is a good idea, it just got me curious ;) (maybe in another PR?)

I didn't know about that, looks neat. The problem here is that some of the deprecation #TODOs are related to decorators already, so not sure how that would work

bolinocroustibat:
> Well done with the refactoring!
> 
> Don't forget to add a changelog entry!

Indeed, fixed!

udata_hydra/routes/__init__.py
bolinocroustibat: It's on purpose, as a reminder of how aiohttp routes can be typed... aiohttp doc is pretty bad so I decided to keep that as a uselful comment.

tests/test_api.py
bolinocroustibat: I think better to keep it as I won't be able to remmeber what was `client_http_method` for and that it will be useless later :) At least for me!

udata_hydra/routes/__init__.py
bolinocroustibat: Done

PR #112, opened 2024-07-30
refactor: factorize DB methods, use DB pool and connection from context

bolinocroustibat:
Related to [issue #1395](https://github.com/datagouv/data.gouv.fr/issues/1395), but not fully closing the issue yet.
Also closes https://github.com/datagouv/hydra/pull/55.

This PR has been split into 3 parts for easier reviewing.
1) **This PR**:  refactoring in order to prepare for new routes (_Note: It had already partly reviewed in previous state of [PR #104](https://github.com/datagouv/hydra/pull/104)_)
2) Adding new route to force a crawl: [PR #104](https://github.com/datagouv/hydra/pull/104)
3) Adding resource statuses and a resource status route: [PR #113](https://github.com/datagouv/hydra/pull/113)

**Details of this PR:**
- Refactor project structure to use DB classes for each DB table, with their factorized DB methods. This was necessary to improve readability and maintainability since routes and cli use common DB queries. This is extremely similar to what has been started in this PR that has been opened for a long time: https://github.com/datagouv/hydra/pull/55. This will also strongly helps to refactor with an ORM later (shttps://github.com/https://github.com/datagouv/hydra/issues/29) and make the whole project more REST/CRUD/MVC standardized.
- Always use DB pool from app context instead of passing it to DB functions through the requests
- Put schemas in dedicated `schemas/` directory
- Add docstrings to some functions to improve documentation

PR #118, opened 2024-07-31
feat: add route for force crawling

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1395 (partly).

This ticket has been split into 5 PRs for easier reviewing:
1) Refactoring DB methods for new routes: [PR #112](https://github.com/datagouv/hydra/pull/112) [✅](https://emojipedia.org/check-mark-button)
2) Refactoring of old routes [PR #117](https://github.com/datagouv/hydra/pull/117) [✅](https://emojipedia.org/check-mark-button)
3) Adding new route to gather informations from a resource in the catalog (ex : url, priority, etc.) [PR #120] [✅](https://emojipedia.org/check-mark-button)
4) **This PR**: Adding new route to force a crawl [PR #118](https://github.com/datagouv/hydra/pull/118)📍
5) Adding resource statuses and a resource status route: [PR #119](https://github.com/datagouv/hydra/pull/119)

**Details of this PR:**
- Create a new POST route `/api/checks/create/` to force a new crawl on a specific `resource_id` or `url`. _Note: no authentification in this PR yet, I think we should do it, but in another PR as this one is already big_
- Create a new related test `tests/test_api.py::test_api_resource_created`


tests/test_api.py
```diff
@@ -94,8 +94,31 @@ async def test_api_get_all_checks(setup_catalog, client, query, fake_check):
     assert second["error"] == "no-can-do"
 
 
-async def test_api_get_resource(db, client, insert_fake_resource):
-    await insert_fake_resource(db)
```
magopian, line 98: The diff in github is a bit weird, but it makes it look like this line is gone? Is that on purpose?

magopian (approved):
Looks good to me :+1:

Do you want to add that to the CHANGELOG? (maybe we need a CI target that automatically checks the CHANGELOG has been updated... or even a tool that automatically populates the CHANGELOG from the commits ? To be discussed ;)

udata_hydra/routes/checks.py
```diff
@@ -22,3 +28,26 @@ async def get_all_checks(request: web.Request) -> web.Response:
     if not data:
         raise web.HTTPNotFound()
     return web.json_response([CheckSchema().dump(dict(r)) for r in data])
+
+
+async def create_check(request: web.Request) -> web.Response:
+    """Create a new check"""
+    try:
+        payload: dict = await request.json()
+        url = payload["url"]
```
Pierlou, line 37: I find it weird to ask for both the resource's URL and id, and I guess it could be error-prone (what happens if you mismatch URL and id?). Wouldn't there be a way to manage with only the id? We could get the URL from the db or data.gouv directly 🤔

tests/test_api.py
```diff
@@ -94,8 +94,31 @@ async def test_api_get_all_checks(setup_catalog, client, query, fake_check):
...
+)
+async def test_api_create_check_wrongly(setup_catalog, post_data, client, fake_check):
+    await fake_check(status=500, error="no-can-do")
+    await fake_check()
+    resp = await client.post("/api/checks/", json=post_data)
+    assert resp.status == 400
+
+
+async def test_api_create_check(setup_catalog, client, fake_check):
+    await fake_check(status=500, error="no-can-do")
```
Pierlou, line 113: I'm guessing this will go eventually and was left as a reminder?

udata_hydra/routes/checks.py
```diff
@@ -22,3 +28,26 @@ async def get_all_checks(request: web.Request) -> web.Response:
...
+    except KeyError as e:
+        raise web.HTTPBadRequest(text=f"Missing key: {e}")
+
+    else:
+        context.monitor().set_status(f'Crawling url "{url}"...')
+
+        async with aiohttp.ClientSession(
+            timeout=None, headers={"user-agent": config.USER_AGENT}
+        ) as session:
+            result = await check_url(url=url, resource_id=resource_id, session=session)
```
Pierlou, line 50: Wondering if in this specific case we'd like the check to have high priority. Currently `check_url` [always sets priority as low](https://github.com/datagouv/hydra/blob/46b73144792392557255a72f3ee63ed88bf2045f/udata_hydra/crawl.py#L288C27-L288C43). Thoughts?

bolinocroustibat:
> Looks good to me 👍
> 
> Do you want to add that to the CHANGELOG? (maybe we need a CI target that automatically checks the CHANGELOG has been updated... or even a tool that automatically populates the CHANGELOG from the commits ? To be discussed ;)

Good point, I forgot. Changed now and also fixed a fe typos in the changelog.

I would be very interested by the second option, to populate the changelog automatically from commit names and descriptions. That would be a very big gain in productivity, and it would force us to do even better commits names and descriptions :)

Pierlou (approved):
Looking forwards to having this in production!

bolinocroustibat:
I will wait for @maudetes to review the whole logic (especially the priority logic) to make sure it's ok, before merging.
So keeping this PR open for a few more days before merging :)

udata_hydra/cli.py
```diff
@@ -144,7 +144,9 @@ async def check_resource(resource_id: str, method: str = "get"):
         log.error("Resource not found in catalog")
         return
     async with aiohttp.ClientSession(timeout=None) as session:
-        await crawl_check_url(url=res[0], resource_id=None, session=session, method=method)
+        await crawl_check_url(
+            url=res[0], resource_id=None, session=session, method=method, worker_priority="low"
```
maudetes, line 148: Actually, since it's quite an equivalent from the API call, we may want to have `worker_priority="high"` as well in the cli.

udata_hydra/crawl.py
```diff
@@ -235,7 +237,12 @@ def has_nice_head(resp) -> bool:
 
 
 async def check_url(
-    url: str, resource_id: str, session, sleep: float = 0, method: str = "head"
+    url: str,
+    resource_id: str,
+    session,
+    sleep: float = 0,
+    method: str = "head",
+    worker_priority: str = "default",
```
maudetes, line 245: Thanks for the renaming, it is clearer indeed!

maudetes (approved):
Thank yo!

I am thinking of future iterations (running the different tasks inline instead of going through a worker, an additional priority level, etc.) but we can start with this version for a starter!

It would be wiser to test this new endpoint for real usage examples.

PR #104, opened 2024-07-26
feat: new route for force crawling

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1395 (partly).

This PR has been split into 3 parts for easier reviewing.
1) Refactoring in order to prepare for new routes: [PR #112](https://github.com/datagouv/hydra/pull/112)
2) **This PR**: Adding new route to force a crawl
3) Adding resource statuses and a resource status route: [PR #113](https://github.com/datagouv/hydra/pull/113)

**Details of this PR:**
- Create a new POST route `/api/checks/create/` to force a new crawl on a specific `resource_id` or `url`. _Note: no authentification in this PR yet, I think we should do it, but in another PR as this one is already big_
- Create a new related test `tests/test_api.py::test_api_resource_created`

PR #119, opened 2024-07-31
feat: update resource statuses in DB and add resource status route

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1395 (partly).

This ticket has been split into 6 PRs for easier reviewing.
1) Refactoring in order to prepare for new routes (_Note: It had already partly reviewed in previous state of [PR #104](https://github.com/datagouv/hydra/pull/104)_)
2) Refactoring of old routes [PR #117](https://github.com/datagouv/hydra/pull/117)
3) Adding new route to gather informations from a resource in the catalog (ex : url, priority, etc.) [PR #120]
4) Adding new route to force a crawl [PR #118](https://github.com/datagouv/hydra/pull/118)
5) Adding authentication logic with API key for some endpoints
6) **This PR**: Adding resource statuses and a resource status route: [PR #119](https://github.com/datagouv/hydra/pull/119)

**Details of this PR:**
- Create `Resource.STATUSES` for different statuses of resources :
```
`None`: "no status, waiting",
"BACKOFF": "backoff period for this domain, will be checked later",
"CRAWLING": "resource URL currently being crawled",
"TO_PROCESS_RESOURCE": "to be processed for change analysis",
"PROCESSING_RESOURCE": "currently being processed for change analysis",
"TO_ANALYSE_CSV": "to be analysed by CSV detective",
"ANALYSING_CSV": "currently being analysed by CSV detective",
"INSERTING_IN_DB": "currently being inserted in DB",
"CONVERTING_TO_PARQUET": "currently being converted to Parquet",
```
- Update `Resource.insert` and `Resource.update_or_insert` methods accordingly
- Update resource status in DB in the crawling and analysis logic according to the status of analysis
- Don't modify the resource status in the DB when updating the priority, in order not to erase the current status
- Create a new GET route `/api/resource/status/` to get the current status of a resource. It is the same as `get_resource` endpoint, but only returns the status of the resource, saving bandwith and processing time (and also uses a more RESTful URL)
- Simplify logic to pick up a batch of resources to be checked: no need to update status to "crawling" anymore when picking up
- Add test to see if checker has been executed for all the different resource statuses cases
- Add test for this new `/api/resource/status/` API route
- There were a confusion in the naming between **crawl** and **check** actions in the code. To be clear, a **check** includes the following steps named:
    - **crawl** the URL (HTTP request to the resource to get the distant resource)
    - **analyse** the data crawled
    .. and other steps.
...so it sounded more consistent and less confusing to change the naming from (`crawl`, `crawl_url` and `check_urls`) to (`check_catalog`, `check_url` and `check_urls`).


README.md
```diff
@@ -281,7 +281,7 @@ UDATA_URI_API_KEY = "example.api.key"
 SENTRY_DSN = "https://{my-sentry-dsn}"
 ```
 
-The webhook integration sends HTTP messages to `udata` when resources are analyzed or checked to fill resources extras.
+The webhook integration sends HTTP messages to `udata` when resources are analysed or checked to fill resources extras.
```
magopian, line 284: American english or queen's english, i won't take sides 😬 (ok, i'd rather use the queen's english, but that's just imho 🤣 )

tests/conftest.py
```diff
@@ -153,14 +155,16 @@ async def db():
...
 @pytest_asyncio.fixture
-async def insert_fake_resource():
-    async def _insert_fake_resource(database) -> None:
-        await database.execute(
-            f"""
-            INSERT INTO catalog (dataset_id, resource_id, url, priority, deleted)
-            VALUES ('{DATASET_ID}', '{RESOURCE_ID}', 'http://dev.local/', True, False)
-            ON CONFLICT (resource_id) DO NOTHING;
-            """
+async def insert_fake_resource(status: Union[str, None] = None):
```
magopian, line 158: I'm not sure the parameters are useful/used in this outer fixture function? I believe they could be simply removed, right?

tests/test_api.py
```diff
@@ -102,6 +102,18 @@ async def test_api_get_resource(db, client, insert_fake_resource):
     data: dict = await resp.json()
     assert data["dataset_id"] == DATASET_ID
     assert data["resource_id"] == RESOURCE_ID
+    assert data["status"] == "TO_CHECK"
+
+
+@pytest.mark.parametrize("resource_status", list(Resource.STATUSES.items()))
+async def test_api_get_resource_status(db, client, insert_fake_resource, resource_status):
```
magopian, line 109: If I understand [pytest parametrization correctly](https://docs.pytest.org/en/7.1.x/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions), you could even:

```suggestion
@pytest.mark.parametrize("resource_status,resource_verbose_status", list(Resource.STATUSES.items()))
async def test_api_get_resource_status(db, client, insert_fake_resource, resource_status, resource_verbose_status):
```
and then use `resource_status` instead of `resource_status[0]`, and `resource_verbose_status` instead of `resource_status[1]`, i think?

udata_hydra/analysis/csv.py
```diff
@@ -74,7 +75,7 @@
 minio_client = MinIOClient()
 
 
-async def notify_udata(check_id: str, table_name: str) -> None:
+async def notify_udata(check_id: int, table_name: str) -> None:
```
magopian, line 78: Wait, it was typed as `str`, and now it's typed as `int`, but there's no change related to that type change anywhere in the code?
I mean, was it just badly typed before, and went undetected? If so, would we want some kind of automatic check/validation to prevent this?

If the typings are meant to help us, but they could be wrong, how useful are they really?

I think this should be adressed in another PR or discussed during a meeting, what do you think?

magopian (approved):
Looks nice! A couple of nits, in case you want to address them ;)

🚢

README.md
bolinocroustibat: Ahah I hesitated, but since previous code was using mostly British English, I decided to go more consistent and follow British English...

tests/conftest.py
bolinocroustibat: You mean the `status` parameter? It's not used yet, but I added it because I might use it later.

tests/conftest.py
```diff
@@ -153,14 +155,16 @@ async def db():
...
+    async def _insert_fake_resource(database, status: Union[str, None] = None):
+        if status is None:
+            status = "TO_CHECK"
```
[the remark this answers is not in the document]
bolinocroustibat: I tried that earlier, but in that case it would mean that a `None` status can be passed to `Resource.insert`, so we would need a slight refacto of Resource.insert, which I didn't want to do in that PR.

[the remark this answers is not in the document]
bolinocroustibat: Screenshot 2024-08-20 at 16 57 07

tests/test_api.py
bolinocroustibat: Good idea, it's easier to read! Thanks! Done

tests/conftest.py
magopian: How would you use this `status` parameter on the fixture itself?

When using a fixture in a test, you use it like that, right?
```python
def test_foo(insert_fake_resource):
    insert_fake_resource("TO_CHECK")
```

But there's no parameter passing here, you can't do
```python
def test_foo(insert_fake_resource("TO_CHECK")):
    insert_fake_resource()
```

So i understand the parameter in the inner function (the helper, the one returned by the fixture), but not the parameter in the fixture function itself.

Am I missing something?

tests/conftest.py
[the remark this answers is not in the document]
magopian: ah yes i understand now, it makes sense 👍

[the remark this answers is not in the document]
bolinocroustibat: Actually, it would be totally OK with `async def _insert_fake_resource(database, status: str = "TO_CHECK"):`

Just fixed!

tests/conftest.py
bolinocroustibat: No, you're right, just fixed! Thanks

udata_hydra/routes/resources.py
```diff
@@ -21,6 +21,29 @@ async def get_resource(request: web.Request) -> web.Response:
...
+        resource_id: str = request.match_info["resource_id"]
+    except Exception as e:
+        raise web.HTTPBadRequest(text=json.dumps({"error": str(e)}))
+
+    resource = await Resource.get(resource_id=resource_id, column_name="status")
+    if not resource:
+        raise web.HTTPNotFound()
+
+    status: str = resource["status"]
+    status_verbose: str = Resource.STATUSES[status]
```
maudetes, line 40: I think we should deal with a status undefined or unknown (for a time at least)?

udata_hydra/crawl.py
maudetes, line 1: There was a usage for the previous `status` column that we should accommodate for.
See : https://github.com/datagouv/hydra/pull/68.

udata_hydra/analysis/csv.py
```diff
@@ -131,6 +133,9 @@ async def analyse_csv(
     table_name = hashlib.md5(url.encode("utf-8")).hexdigest()
     timer.mark("download-file")
 
+    # Update resource status to TO_ANALYSE
+    await Resource.update(resource_id, {"status": "TO_ANALYSE"})
```
maudetes, line 137: I think this should happen before `analyse_csv` job even starts.

udata_hydra/cli.py
maudetes, line 1: I think `check-resource` command does not work anymore (probably due to previous refactoring)

maudetes (commented):
Thank you for this status update! This is a WIP review with some pointers

udata_hydra/analysis/csv.py
bolinocroustibat: Yes but I think it would make sense not to change the status before `config.CSV_ANALYSIS` is checked to be true. I moved it right after.

udata_hydra/crawl.py
```diff
@@ -294,10 +308,18 @@ async def check_url(
                 },
             )
 
+            # Update resource status to TO_PROCESS_RESOURCE
+            await Resource.update(resource_id, data={"status": "TO_PROCESS_RESOURCE"})
+
             queue.enqueue(process_resource, check_id, is_first_check, _priority=worker_priority)
 
+            # Reset resource status to None so that it's not forbidden to be checked again
+            await Resource.update(resource_id=resource_id, data={"status": None})
```
maudetes, line 317: I think setting resource status to `None` here makes it seems like the process is one, but it continues with `process_resource`.

udata_hydra/db/resource.py
```diff
@@ -1,9 +1,24 @@
+from typing import Optional
+
 from udata_hydra import context
 
 
 class Resource:
     """Represents a resource in the "catalog" DB table"""
 
+    STATUSES = {
```
maudetes, line 9: I think adding a `STALE` with a verbose description (as in not currently being checked) could make it clearer than leaving it to None?

maudetes (commented):
Thank you! It's quite satisfying to run it on the catalog and see the status evolve! :star_struck:

udata_hydra/crawl.py
bolinocroustibat: Done

udata_hydra/cli.py
bolinocroustibat: Thanks for spotting, fixed!

udata_hydra/db/resource.py
Pierlou:
```suggestion
    STATUSES = {
        "CRAWLING_URL": "resource URL currently being crawled",
        "CRAWLED_URL": "resource URL crawl finished",
    }
```
I'm not sure why some tasks have a "finished" status while others don't

udata_hydra/routes/resources.py
bolinocroustibat: Done!

udata_hydra/db/resource.py
bolinocroustibat: @maudetes added a `None: "no status, waiting"` status

udata_hydra/analysis/resource.py
```diff
@@ -101,10 +105,19 @@ async def process_resource(check_id: int, is_first_check: bool) -> None:
     if change_status == Change.HAS_CHANGED:
         await store_last_modified_date(change_payload or {}, check_id)
 
+    # Update resource status to PROCESSED_RESOURCE
+    await Resource.update(resource_id, data={"status": "PROCESSED_RESOURCE"})
+
     analysis_results = {**dl_analysis, **(change_payload or {})}
+
     if change_status == Change.HAS_CHANGED or is_first_check:
         if is_tabular and tmp_file:
```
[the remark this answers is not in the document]
bolinocroustibat: Understood, changed, thanks!

udata_hydra/db/resource.py
bolinocroustibat: @Pierlou the only status which had a "finished" status was CRAWLED, on purpose, because several stuff could happen between CRAWLED and TO_PROCESS_RESOURCE, but actually after reading more I was confused with the crawling step. So I removed it. Thanks!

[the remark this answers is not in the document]
bolinocroustibat: See conversation here after

udata_hydra/routes/resources.py
```diff
@@ -21,6 +21,29 @@ async def get_resource(request: web.Request) -> web.Response:
...
+
+    return web.json_response(
```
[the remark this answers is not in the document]
bolinocroustibat: Good idea! WIP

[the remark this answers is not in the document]
bolinocroustibat: Done

udata_hydra/routes/resources.py
```diff
@@ -21,6 +23,36 @@ async def get_resource(request: web.Request) -> web.Response:
...
+        raise web.HTTPBadRequest(text=json.dumps({"error": str(e)}))
+
+    resource = await Resource.get(resource_id=resource_id, column_name="status")
+    if not resource:
+        raise web.HTTPNotFound()
+
+    status: Optional[str] = resource["status"]
+    status_verbose: str = Resource.STATUSES[status]
+
+    latest_check_url: str = f"{request.scheme}://{request.host}{str(request.app.router["get-latest-check"].url_for())}?resource_id={resource_id}"
```
maudetes, line 44: Woops, quotes used in f-string don't work with python3.9

maudetes (approved):
Thank you a lot for adding status and the refactor along the way!

Can't wait to have it in production :rocket:

udata_hydra/routes/resources.py
bolinocroustibat: Well spotted, fixed!

PR #122, opened 2024-07-31
feat: add tests coverage

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/121.

PR #123, opened 2024-08-01
tests: add tests

bolinocroustibat:
Since we added tests coverage (https://github.com/datagouv/hydra/pull/122), I saw that some API lines were not covered by tests.

This PR add some tests coverage, and slightly refactor API tests.


magopian (approved):
Nice! I like the usage of `RESOURCE_ID` now, makes more sense.
🚢

PR #125, opened 2024-08-01
clean: minor docs and linting fixes

bolinocroustibat:
Very minor fixes on the go:
- Add a docstring
- Fix a comment typo
- Fix pre-commit with updated Ruff args

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,6 +14,6 @@ repos:
     hooks:
       # Run the linter
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix, --select, I]
+        args: [--fix, --exit-non-zero-on-fix, --extend-select, I]
       # Run the formatter
       - id: ruff-format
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -14,6 +14,6 @@ repos:
     hooks:
       # Run the linter
       - id: ruff
-        args: [--fix, --exit-non-zero-on-fix, --select, I]
+        args: [--fix, --exit-non-zero-on-fix, --extend-select, I]
```
magopian, line 17: If you add a `ruff.toml` file like the one on [udata](https://github.com/opendatateam/udata/pull/3118/files#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989), you shouldn't need to specify the options like that.

magopian (approved):
Nice 👍 
Adding a [ruff.toml](https://github.com/opendatateam/udata/pull/3118/files#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989) should allow us to not specify the `extend-select`, just FYI (i've tested locally on udata, and it is indeed taken into account).

PR #126, opened 2024-08-01
fix: fix CLI typing

bolinocroustibat:
Quick fix of types in CLI:
- Since some params are optional OR the other one, maybe that's why Union typing was failing - when a default argument follows a non-default argument
- By the way, fixed return a type hint that was not compatible with Python 3.9

Detail :

Les hint types font crash certaines commandes CLI de mon côté, notamment analyse-csv, tu confirmes que ça fonctionne chez toi ?

Par exemple pour `poetry run udata-hydra analyse-csv --url https://object.files.data.gouv.fr/meteofrance/data/synchro_ftp/BASE/HOR/H_03_1970-1979.csv.gz` :
Si je laisse

```
@cli(name="analyse-csv")
async def analyse_csv_cli(
    check_id: Union[int, None] = None, url: Union[str, None] = None, debug_insert: bool = False
):
```
ça plante (udata-hydra analyse-csv: error: argument --url/-u: invalid Optional value) (pareil avec Optional)
Mais si je mets

`... url: str = None, ...`

ça passe


Pierlou (approved):
LGTM, is it expected or an unwanted behaviour of `minicli`?

Pierlou:
Following https://github.com/pyrates/minicli/pull/13 we can simply upgrade `minicli` to 0.5.3

Pierlou (approved):

PR #130, opened 2024-08-14
feat: add authentication

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1417:

- Add basic authentication via admin token using a new auth logic in `/utils/auth.py`, for all `POST`/`PUT`/`DELETE` endpoints (= not using `GET` method)
- Refactor tests to include tests cases when no token or invalid token for each `POST`/`PUT`/`DELETE`

_Minor:_
- Sort dependencies in `pyproject.toml` alphabetically


pyproject.toml
```diff
@@ -17,27 +17,28 @@ readme = "README.md"
 
 [tool.poetry.dependencies]
 python = "^3.9"
+aiocontextvars = "^0.2.2"
 aiohttp = "^3.10.3"
+aiohttp-tokenauth = "^0.0.2"
```
maudetes, line 22: I think we may not want to install this lib with last modification 5 years ago?
I would rather copy the logic in hydra directly since the usage looks clean and the code simple enough?
What do you think?

maudetes (commented):
Thank you for this PR!

pyproject.toml
bolinocroustibat: That's a very good idea, it will greatly simplifies the code, make it more readable and improve maintanability.
Just commited changes to do that.

maudetes (approved):
Nice, looks quite simple and easily extendable this way!

PR #132, opened 2024-08-16
fix: fully RESTful endpoints

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/160

Refactor several routes so that the API gets very close to full REST standards:

- New routes have been implemented this way:
    - [x] `GET` to `/api/resources` duplicated a a route with no resource payload in the body, but as a `GET` on `/api/resources/{resource_id}`
    - [x] `PUT` to `/api/resources` duplicated as a `PUT` to /`api/resources/{resource_id}` with the same payload
    - [x] `DELETE` to `/api/resources` duplicated as a route with no payload, but as a `DELETE` on `/api/resources/{resource_id}`

- Some routes responses have been changed that way:
    - [x] `POST` to `/api/resources` answers a HTTP Created (201) instead of HTTP OK (200) with "created" as content
    - [x] `PUT` to /`api/resources/{resource_id}` answers a HTTP OK (200) with empty body instead of HTTP OK (200) with "updated" as content
    - [x] `DELETE` to /`api/resources/{resource_id}` should answer a HTTP OK with no content (204) with empty body instead of HTTP OK (200) with "deleted" as content
    - [x] `POST` to `/api/checks` should answer a HTTP Created (201) with the created object, instead of HTTP OK (200)

- The following have been moved to separate, specific places, while still being active, to preserve retro-compatibility while making a clear cut between current and legacy code:
    - Legacy routes URLs: `legacy_routes` and `legacy_routes_params` in [`udata_hydra/routes/__init__.py`](https://github.com/datagouv/hydra/pull/132/files#diff-e5ce70c8c6e9d9d078789d97ab169134b487c04923fa2b5b8bc584edee548e06)
    - Legacy API routes logic: [`udata_hydra/routes/resources_legacy.py`](https://github.com/datagouv/hydra/pull/132/files#diff-fafc7cf4d53a0c0beabd06c3a7dc5b6b5343e69d020f399c591658ca9cf8f150)
    - Legacy API routes tests: [`tests/test_api_legacy.py`](https://github.com/datagouv/hydra/pull/132/files#diff-22420b21d89ad900912688af4fed0206b66f6789c8438218eed67f8871cf0865)
    
- Add a few tests cases on the go. Code coverage is 88%.


magopian (commented):
Nice, looks way more RESTful and standard like that 👍 
I'm curious about the URLs, are those endpoints used anywhere, which would need updating? I'm not familiar with the hydra project, so I'm not sure I'm competent enough to approve the PR 🤷

bolinocroustibat:
> Nice, looks way more RESTful and standard like that 👍 I'm curious about the URLs, are those endpoints used anywhere, which would need updating? I'm not familiar with the hydra project, so I'm not sure I'm competent enough to approve the PR 🤷

They are used in udata if I'm not wrong - but I kept the legacy endpoint for now so that shouldn't impact anything yet.
We would nee to look on udata code to update the calls to hydra at some point indeed.

magopian:
Doesn't [this](https://github.com/datagouv/hydra/pull/132/files#diff-19047ee66282842315b0e739aee81913ddc50e6341f1dd68df3108b9fcb379ddR17) break the previous routes (which were looking for the `resource_id` in the parameters), and replace them with routes with the `resource_id` in the path?

bolinocroustibat:
> Doesn't [this](https://github.com/datagouv/hydra/pull/132/files#diff-19047ee66282842315b0e739aee81913ddc50e6341f1dd68df3108b9fcb379ddR17) break the previous routes (which were looking for the `resource_id` in the parameters), and replace them with routes with the `resource_id` in the path?

Indeed it does. I will put that PR away for now, we'll do that later. Let's change step by step :)

bolinocroustibat:
> Doesn't [this](https://github.com/datagouv/hydra/pull/132/files#diff-19047ee66282842315b0e739aee81913ddc50e6341f1dd68df3108b9fcb379ddR17) break the previous routes (which were looking for the `resource_id` in the parameters), and replace them with routes with the `resource_id` in the path?

For information since your last review, the PR has been refactored to be fully compatible with old udata routes, which all stand in `***_legacy` files or objects, so that it'll be easy to read and to remove them later.

tests/test_api.py
```diff
@@ -332,43 +332,34 @@ async def test_api_update_resource(client, route, api_headers, api_headers_wrong
     }
 
     # Test API call with no token
-    resp = await client_http_method(path=route["url"], headers=None, json=payload)
+    resp = await client.put(path=f"/api/resources/{RESOURCE_ID}", headers=None, json=payload)
```
magopian, line 335: Same here, just above in the diff, on line [311](https://github.com/datagouv/hydra/pull/132/files#diff-3451cc474d5698222168f66a293339a6bd296ee2c56f56409866bbae99917a79R311) you used the url WITH the final `/`, and here you use the same url WITHOUT the final `/`.

As long as it's on purpose i'm fine with it. I wonder if this would warrant a special parametrization, or maybe some tests dedicated to just that (using url with and without the final `/` just to make sure both work exactly the same).

tests/test_api.py
```diff
@@ -393,59 +384,44 @@ async def test_api_update_resource_url_since_load_catalog(
...
-async def test_api_delete_resource(client, route, api_headers, api_headers_wrong_token):
-    # TODO: can be removed once we don't use legacy route anymore
-    client_http_method: Callable = getattr(client, route["method"])
-
-    # Test invalid DELETE data
-    stupid_delete_data: dict = {"stupid": "stupid"}
-    resp = await client_http_method(path=route["url"], headers=api_headers, json=stupid_delete_data)
+async def test_api_delete_resource(client, api_headers, api_headers_wrong_token):
+    # Test invalid resource_id
+    resp = await client.delete(path="/api/resources/STUPID-ID", headers=api_headers)
```
magopian, line 398: I'm assuming this PR will need https://github.com/datagouv/data.gouv.fr/issues/1426 merged first, and this PR actually removes the legacy routes? (and thus removes the legacy test code).

udata_hydra/routes/resources.py
```diff
@@ -1,20 +1,25 @@
...
 async def get_resource(request: web.Request) -> web.Response:
     """Endpoint to get a resource from the DB
     Respond with a 200 status code and a JSON body with the resource data
     If resource is not found, respond with a 404 status code
     """
-    [resource_id] = get_request_params(request, params_names=["resource_id"])
+
+    try:
+        resource_id = str(uuid.UUID(request.match_info["resource_id"]))
+    except Exception as e:
```
magopian, line 20: It's a `ValueError` that's raised, right? Would we want to specify the kind of exception?

udata_hydra/routes/resources.py
```diff
@@ -29,7 +34,7 @@ async def get_resource_status(request: web.Request) -> web.Response:
     If resource is not found, respond with a 404 status code
     """
     try:
-        resource_id: str = request.match_info["resource_id"]
+        resource_id = str(uuid.UUID(request.match_info["resource_id"]))
     except Exception as e:
```
magopian, line 38: I see you've reused the same `except Exception as e` above, I understand now ;) Code consistency!

magopian (approved):
I like that you moved the legacy routes in another file, makes it easier to get rid of them once they're fully deprecated 💯 
This looks good to me, with a few minor nits and remarks, feel free to ignore them!

tests/test_api.py
```diff
@@ -250,66 +268,48 @@ async def test_api_get_resource_status(
...
 async def test_api_create_resource(
-    client, route, api_headers, api_headers_wrong_token, udata_resource_payload
+    client, api_headers, api_headers_wrong_token, udata_resource_payload
 ):
-    # TODO: can be removed once we don't use legacy route anymore
-    client_http_method: Callable = getattr(client, route["method"])
-
     # Test API call with no token
-    resp = await client_http_method(path=route["url"], headers=None, json=udata_resource_payload)
+    resp = await client.post(path="/api/resources/", headers=None, json=udata_resource_payload)
```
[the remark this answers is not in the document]
bolinocroustibat: Yes, it's on purpose. In a previous merged PR, we added both the trailing slash option and no trailing slash option for each route (see routes/__init__.py) in oder not not have to thin about this anymore.

Since it's now done, I refactored the tests to make the calls with no trailing slash in this present PR, which is a bit more standard to me.

udata_hydra/routes/resources.py
bolinocroustibat: It can be either a `ValueError`, either an error because it's not a valid UUID.
Could be indeed nice to go further and only catch those two specific exceptions.

tests/test_api.py
bolinocroustibat: No, this PR keeps the legacy route (see https://github.com/datagouv/hydra/pull/132/files#diff-fafc7cf4d53a0c0beabd06c3a7dc5b6b5343e69d020f399c591658ca9cf8f150 and https://github.com/datagouv/hydra/pull/132/files#diff-22420b21d89ad900912688af4fed0206b66f6789c8438218eed67f8871cf0865 for the related tests)

tests/test_api.py
bolinocroustibat: Same here, it's on purpose but since a previous merged PR, both work. See https://github.com/datagouv/hydra/pull/132/files#diff-e5ce70c8c6e9d9d078789d97ab169134b487c04923fa2b5b8bc584edee548e06R22

tests/test_api.py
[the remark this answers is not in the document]
bolinocroustibat: See https://github.com/datagouv/hydra/pull/132/files#diff-e5ce70c8c6e9d9d078789d97ab169134b487c04923fa2b5b8bc584edee548e06R22

udata_hydra/routes/resources.py
bolinocroustibat: I'll keep that for a future PR though :)

udata_hydra/routes/__init__.py
```diff
@@ -42,14 +48,11 @@ def generate_routes(
     (web.get, "/api/checks/all", get_all_checks, None),
     (web.post, "/api/checks", create_check, None),
     # Routes for resources
-    (web.get, "/api/resources", get_resource, None),
+    (web.get, "/api/resources/{resource_id}", get_resource, None),
     (web.get, "/api/resources/{resource_id}/status", get_resource_status, None),
     (web.post, "/api/resources", create_resource, None),
```
maudetes, line 53: I wonder if we should update the payload on this route at some point, but would probably be done in a future PR if needed.

maudetes (approved):
:clap:
Thank you for the clear separation between legacy routes and new routes!

tests/test_api.py
```diff
@@ -238,9 +246,19 @@ async def test_api_get_resource(setup_catalog, client):
...
+    # Test invalid resource_id
+    resp = await client.get(path="/api/resources/STUPID-ID/status")
+    assert resp.status == 400
+
+    # Test non existing resource_id
+    resp = await client.get(path=f"/api/resources/{NOT_EXISTING_RESOURCE_ID}/status")
+    assert resp.status == 404
+
+    # Test existing resource
+    resp = await client.get(f"/api/resources/{RESOURCE_ID}/status")
```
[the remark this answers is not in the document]
bolinocroustibat: No it's not the same endpoint, one is /api/resources/{NOT_EXISTING_RESOURCE_ID}`, the other one is /api/resources/{NOT_EXISTING_RESOURCE_ID}/status`

PR #134, opened 2024-08-21
clean: fix type hints

bolinocroustibat:
Minor PR.

- Fix a few type hints
- Better type hints (use of <3.10 style `Optional`instead of `Union[...,None]`)
- Add comment and modify doctrings


magopian (approved):
Looks way better, thanks, love it!

🚢

PR #138, opened 2024-08-21
feat: load pyproject.toml in config

bolinocroustibat:
Simplify the loading of `APP_NAME` and `APP_VERSION` from `pyproject.toml`: load them in config, to be later used by Sentry.


magopian (approved):
Way nicer indeed, thanks! 🚢

PR #139, opened 2024-08-21
Document the settings needed in udata to publish resource events

magopian:
This follows #130, and the associated [PR in udata](https://github.com/opendatateam/udata/pull/3130).


maudetes (approved):
:ok_hand:

PR #140, opened 2024-08-22
fix: fix check-resource CLI command bug

bolinocroustibat:
Fix `check_resource` CLI command.


maudetes (approved):
Thank you for the fix :pray:

PR #142, opened 2024-08-26
build: update csv-detective to 0.7.2.dev824

bolinocroustibat:
Update `csv-detective` to [0.7.2.dev824](https://pypi.org/project/csv-detective/#history) which doesn't include yanked version of `requests` anymore.

Also quickly locally tested with other package managers than Poetry (which are more careful than Poetry which just outputs warnings for the yanked), works fine with no error anymore.


Pierlou (approved):
🚀

PR #143, opened 2024-08-26
feat: PEP-621 standard Python project metadata

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/145 by:

- Refactor `pyproject.toml` to become a package-manager agnostic project file
- Generate locked dependencies files as `requirements.txt` and `requirements-dev.txt` files from `pyproject.toml` Python project file
_Question: do we want to include dependencies hashes in the lock dependencies files? Can be done with `pip-compile --generate-hashes`_
- Refactor CI not to use Poetry files
- Change version number in `pyproject.toml`so that when CI adds dev version, it is a PEP440 compatible
- Update README and documentation to explain how to install and generate lock files


.circleci/config.yml
```diff
@@ -50,12 +53,13 @@ jobs:
...
-          # ruff check --select I . : check linting and imports sorting without fixing (to fix, use --fix)
-          # ruff format --check . : check code formatting without fixing (to fix, remove --check)
+          name: Lint code, format code and sort imports
           command: |
-            poetry run ruff check --select I .
-            poetry run ruff format --check .
+            source .venv/bin/activate
+            # Check linting and imports sorting (to fix, add --fix)
+            ruff check
+            # Check code formatting (to fix, add --check)
```
magopian, line 61: Nit: his comment looks weird, it says to add `--check`, but the line below does already have the `--check`

README.md
```diff
@@ -22,26 +22,45 @@ The hydra crawler is one of the components of the architecture. It will check if
...
 ## Dependencies
 
+### System
+
 This project uses `libmagic`, which needs to be installed on your system, eg:
+`brew install libmagic` on MacOS, or `sudo apt-get install libmagic-dev` on Linux.
+
+### Python
+
+This project uses Python 3.9.
```
magopian, line 32: Nit: I believe i used python 3.12 at least for a few reviews and for the couple of PRs I created, and it seemed to work correctly. Would we want something like:

```suggestion
This project uses Python >=3.9.
```

README.md
```diff
@@ -22,26 +22,45 @@ The hydra crawler is one of the components of the architecture. It will check if
...
 
-`brew install libmagic` on MacOS, or `sudo apt-get install libmagic-dev` on linux.
+Project dependencies are listed in `pyproject.toml`, while dependencies are locked in `requirements.txt` (for production only deps) and in `requirements-dev.txt` files.
+
+To install the exact same environment locally including the dev dependencies, use the lock file:
+`pip install -r requirements-dev.txt`
+...or `pip-sync requirements-dev.txt` with [pip-tools](https://pip-tools.readthedocs.io/en/stable/).
+
+To update the lock files, you can use any modern Python package manager (except Poetry) like [pip-tools](https://pip-tools.readthedocs.io/en/stable/), [PDM](https://pdm.fming.dev/) or [uv](https://uv.readthedocs.io/en/latest/), while defining `requirement.txt` and `requirement-dev.txt` as the output lock files.
+With [pip-tools](https://pip-tools.readthedocs.io/en/stable/), the command are `pip-compile` for requirements.txt, and `pip-compile --extra dev -o requirements-dev.txt` for requirements-dev.txt.
```
magopian, line 41: Nit: small typo: `the commandS are...`

README.md
```diff
@@ -22,26 +22,45 @@ The hydra crawler is one of the components of the architecture. It will check if
...
 
 ## CLI
 
 ### Create database structure
 
-Install udata-hydra dependencies and cli.
-`poetry install`
+Create a Python 3.9 virtual environment and activate it:
+`python3 -m venv .venv && source .venv/bin/activate`
```
magopian, line 48: If we want to pin the usage on python3.9, I guess we need the command to be
```suggestion
`python3.9 -m venv .venv && source .venv/bin/activate`
```

pyproject.toml
```diff
@@ -77,14 +71,15 @@ markers = [
 addopts = "--cov=udata_hydra/ --cov-report term-missing"
 
 [tool.ruff]
-lint = { extend-select = ["I"], ignore = ["F401"] } # ["I"] is to also sort imports with an isort rule # TODO: remove ignore F401 later
+lint = { extend-select = ["I"], ignore = ["F401"] }
+# ["I"] is to also sort imports with an isort rule
+# TODO: remove ignore F401 later
 line-length = 100
```
magopian, line 77: FYI, on udata we have a `ruff.toml` file which does the job quite well imho:

```toml
line-length = 100

[lint]
extend-select = ["I"] # also sort imports with an isort rule
```

magopian (approved):
This looks NICE! Disclaimer, I didn't double check the requirements nor the install of the dependencies, I can do it if needed, if we don't trust the CI ;)

pyproject.toml
bolinocroustibat: I still have some issues with imports in `__init__.py` files which are flagged by Ruff if I don't ignore the F401 rule. If you have suggestions on how to solve that I'm all ears!

.circleci/config.yml
bolinocroustibat: Changed!

pyproject.toml
magopian: The [ruff docs](https://docs.astral.sh/ruff/rules/unused-import/) have a few ideas to fix it. If you still feel uncomfortable about those fixes, I guess there are two other options left (if you want to get rid of the `ignore = ["F401"]` rule):
1. explicitly disable it with an inline comment `# noqa: F401`
2. explicitly disable it for the whole module `# ruff: noqa: F401`

I'm fine with keeping the rule ignore in the config though, so it's your call ;)

bolinocroustibat: Great, thank you for the info. I changed the behavoir. t's indeed better!

maudetes (commented):
Thank you for this standardisation! :rocket:

PR #152, opened 2024-08-30
fix: fix project info loading

bolinocroustibat:


maudetes (approved):
Thanks for the quickfix! May be improved later on :+1:

PR #154, opened 2024-08-30
Fix crawling status

maudetes:
Follows https://github.com/datagouv/hydra/pull/119

Use CRAWLING_URL instead of CRAWLING:
https://github.com/datagouv/hydra/blob/9b5ec880595de505f4a48b89637bfbde01e7f79b/udata_hydra/db/resource.py#L12

Fix https://errors.data.gouv.fr/organizations/sentry/issues/145139/?project=21&referrer=issue-stream


bolinocroustibat (approved):
Well spotted, thanks, LGTM!

PR #155, opened 2024-08-30
Add trailing slash on status route for coherence with others

maudetes:
Follows https://github.com/datagouv/hydra/pull/119

We would actually like to have a redirection instead of a 404 on trailing slash or not, but for now let's keep it coherent with other routes at least.


bolinocroustibat:
Thanks for spotting. The trailing slash management in aiohttp routes is really not ideal, there should be a way for both to work. For now thanks for the fix, LGTM!

bolinocroustibat (approved):

PR #156, opened 2024-09-02
clean: simplify `save_as_parquet` method, and fix type not compatible with Python 3.9

bolinocroustibat:
Minor PR

- simplify `save_as_parquet` method
- fix type not compatible with Python 3.9


tests/test_parquet_export.py
```diff
@@ -36,7 +38,6 @@ async def test_parquet_conversion(
         records=generate_records(file_path, inspection, columns),
         columns=columns,
         output_name=None,
```
Pierlou, line 40: Should help tests pass
```suggestion
        output_filename=None,
```

Pierlou (changes requested):

Pierlou (approved):

PR #158, opened 2024-09-05
fix: allow routes with no trailing slashes

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/159.

- Allow all routes URLs to accept with or without trailing slashes.
It's a bit of an ugly way to do it, as I couldn't find a straighforward way to do it with aiohttp. See https://stackoverflow.com/questions/74779274/ignore-trailing-slashes-in-aiohttp-routes for example.

- Update `drop_dbs` CLI method to drop tables with CASCADE, in case some other table coming from other branch and depending on the current DB tables still have a contraint in DB.


Pierlou:
My previous comment doesn't work as it is due to the different request types 😅 so maybe this:
```
routes_params= [
    ("/api/checks/latest", get_latest_check, "get"),
    ("/api/resources", delete_resource, "delete"),
    ...
]
routes = [
    getattr(web, method)(process(r), f)
    for route, function, method in routes_params
    for process in [lambda x: x, lambda x: x+"/"]
]
```
There must be a more elegant way to achieve this 🤔

bolinocroustibat:
@Pierlou Yep, can't be uglier :) We also need not to forget the name if the route, if any. Will look into it

bolinocroustibat:
@Pierlou Cleaned. Could you review again? Tx!

Pierlou (approved):
🚀

PR #163, opened 2024-09-10
Improve timing of checks depending on resources last modification date

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1312.

- Removes SQL query wrapper `select_rows_based_on_query` which was used to create a temporary table and temporarily mark resources being crawled as "crawling".
Since we update the resource status in the check code elsewhere, it should not be necessary anymore to have this temporary table.
**On review, please double-check the validity of this removal**

- Add `CHECK_DELAY_DEFAULT` config var: time after which a resource must be check if we don't know about its last modification date (replaces the previous `SINCE` config var)

- Add `CHECK_DELAYS` config var: list of delays (postgresql interval syntax) between two checks, if the resource has not been modified

- Update the `select_batch_resources_to_check` method with the following logic:
    1. First adds resources with priority=True to the batch (_this doesn't change_)
    2. Then adds resources without last check (=with no check, never have been checked) to the batch (_this doesn't change_)
    3. Then, if the total number of selected resources is still less than the batch size:
        3.1. add resources which don't have at least two checks yet, and the last check is older than `CHECK_DELAY_DEFAULT`
        3.2. add resources which at least one the their two most recent last checks has an unknown modified date (no `detected_last_modified_at`, so we cannot compare with another check), and the last check is older than `CHECK_DELAY_DEFAULT`
        3.3. add ressources to which their two most recent last checks have changed, and the last check is older than `CHECK_DELAYS[0]` (this is to re-check resources that we know it has changed)
        3.4. add ressources to which their last two checks have not changed, the two checks have done between two delays in `CHECK_DELAYS`, and the last check is older than the same delay in `CHECK_DELAYS` (this is in order to avoid checking too often the same resource which doesn't change) 

- Adapt the way we count `pending_checks` and `fresh_checks` in the `/api/status/crawler` endpoint response:
    - `pending_checks` are the number of resources with no check, plus the number with outdated checks (outdated checks are now counted differently, depending on the delays)
     - `fresh_checks` are the number of resources with a check, minus the number with outdated checks (outdated checks are now counted differently, depending on the delays)


magopian (approved):
I don't understand what was the use of the temporary table, was it to make sure that it's threadsafe?

Other than a small nit, this looks good to me ;)

magopian:
Also, didn't investigate why, but it seems you've got a failing test.

udata_hydra/config_default.toml
```diff
@@ -42,13 +43,17 @@ LARGE_RESOURCES_EXCEPTIONS = [
...
-# crawl batch size, beware of open file limits
+# check batch size, beware of open file limits
 # ⚠️ do not exceed MAX_POOL_SIZE
 BATCH_SIZE = 40
-# crawl url if last check is older than
-SINCE = "1w"
+
+# check resource if last check is older than
+CHECK_DELAYS = ["12 hours", "1 day", "7 days", "30 days"] # must be a list of strings, postgres interval syntax
+CHECK_DELAY_DEFAULT = "1w"
```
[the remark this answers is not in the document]
bolinocroustibat: Good point! `CHECK_DELAY_DEFAULT` also needs to have postgres interval syntax indeed. Fixed, thanks!

maudetes (commented):
Thinking without having dived in the SQL query, I have some broad interrogation :

* do you think we could add a `next_check_at` on the `checks` that we would fill when saving the checks, having the logic being set at that time ?
* or do you think we could use a SQL function to build this `next_check_at` in which the complexity may be hidden for a simpler query on the select batch query ? and easily reused

I'm thinking from a debugging perspective when we'll want to understand or easily compute supposed checking dates, we may want to make it easier to debug, keep history, and to know when or why a resource have been checked (or not) at that time?

udata_hydra/crawl/process_check_data.py
```diff
@@ -31,10 +31,36 @@ async def process_check_data(check_data: dict) -> Tuple[Record, bool]:
...
+            )
+            if previous_delay > timedelta(hours=config.CHECK_DELAYS[-1]):
+                next_check = datetime.now(timezone.utc) + timedelta(hours=config.CHECK_DELAYS[-1])
+            else:
+                for d in config.CHECK_DELAYS:
+                    if previous_delay <= timedelta(hours=d):
+                        next_check = datetime.now(timezone.utc) + timedelta(hours=d)
+                        break
+        check_data["next_check"] = next_check
+
```
magopian, line 59: I had a hard time understanding what that was doing, maybe it's just me...

Would that be clearer like so (I may be missing some detail or edge cases):

```suggestion
            # Resource has not been modified since last check: 
            # Re-check it after a delay just greater than the time elapsed since the last check.
            since_last_check: timedelta = datetime.now(timezone.utc) - datetime.fromisoformat(
                last_check["created_at"]
            )
            now: datetime = datetime.now(timezone.utc)
            # At most, re-check after the greatest delay.
            next_check = now + timedelta(hours=config.CHECK_DELAYS[-1])
            # Should we re-check it sooner? Loop through the CHECK_DELAYS from longest to shortest.
            for delay in reversed(config.CHECK_DELAYS):
                delay_delta: timedelta = timedelta(hours=delay)
                if since_last_check <= delay_delta:
                    next_check = now + delay_delta
                else:
                    # Previous delay was the last one greater than `since_last_check`, keep it.
                    break
        check_data["next_check"] = next_check
```

I realize my rewrite is slightly longer, it feels clearer to me, but it could very well be because i just (re)wrote it, feel free to simply dismiss this.

udata_hydra/routes/status.py
```diff
@@ -30,41 +27,84 @@ async def get_status_counts(request: web.Request) -> dict[str, int]:
...
+                    AND MAX(recent_checks.created_at) >= NOW() AT TIME ZONE 'UTC' - INTERVAL '{config.CHECK_DELAYS[i + 1]}'
+                    AND MIN(recent_checks.created_at) < MAX(recent_checks.created_at) - INTERVAL '{config.CHECK_DELAYS[i]}'
+                    AND MIN(recent_checks.created_at) >= MAX(recent_checks.created_at) - INTERVAL '{config.CHECK_DELAYS[i + 1]}'
+                )
+        """
+
+    q_end = f"""
+                OR MAX(recent_checks.created_at) < NOW() AT TIME ZONE 'UTC' - INTERVAL '{config.CHECK_DELAYS[-1]}'
+            )
+        ) AS subquery;
```
magopian, line 91: I'm afraid I'd be unable to validate that this works properly without the tests to back it up, and I'm not even sure I understand the various cases explained in the comment above the query, I'm sorry :(

magopian (approved):
As far as I can tell, this is very nice and impressive, pretty complex piece of code!

I believe there's a typo that breaks the behavior that you may want to fix, other than that, I'm afraid I can't tell much more, in particular about the SQL query 😱 

I'll approve this PR and let you decide if you want to go forward with this PR as is, or if you want to add tests to validate that the SQL query does what it's meant to do? Maybe there area already some that were there previously?

udata_hydra/crawl/process_check_data.py
```diff
@@ -31,10 +31,36 @@ async def process_check_data(check_data: dict) -> Tuple[Record, bool]:
 
     is_first_check: bool = last_check is None
 
+    # Calculate next check date
+    if is_first_check:
+        next_check = datetime.now(timezone.utc) + timedelta(hours=config.CHECK_DELAY_DEFAULT)
+
+    else:
+        if has_check_changed:
```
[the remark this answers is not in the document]
bolinocroustibat: It is totally a typo! Thank you for noticing :)

udata_hydra/routes/status.py
bolinocroustibat: Damned, I completely forgot about this file, that's actually a mistake! It's an old commit that shouldn't be there. The idea is to avoid this previous way of doing it with a overly complicated and unmaintainable SQL query.
Will remove a write a bit

bolinocroustibat: Thanks for noticing!

bolinocroustibat: Fixed!

udata_hydra/crawl/process_check_data.py
bolinocroustibat: Very nice, thanks, although I wonder if I prefer going from the shorter delay to the longest... As the shortest should be the default (more safe), and going to longer delays should be considered as "fallbacks", if that make sense.

I might make a mix your solution with mine though!

udata_hydra/analysis/resource.py
```diff
@@ -32,7 +33,7 @@ class Change(Enum):
 
 
 async def analyse_resource(
-    check_id: int, is_first_check: bool, force_analysis: bool = False
+    check_id: int, last_check: dict | None, force_analysis: bool = False
```
maudetes, line 36: Since we're using last_check here, we could actually forward it to the different detect change function that fetch it.
However, since it's already quite a PR, let's do that later on

udata_hydra/crawl/select_batch.py
```diff
@@ -66,23 +65,22 @@ async def select_batch_resources_to_check() -> list[Record]:
...
             limit = config.BATCH_SIZE - len(to_check)
             q = f"""
             SELECT * FROM (
                 SELECT catalog.url, dataset_id, catalog.resource_id
                 FROM catalog, checks
                 WHERE catalog.last_check IS NOT NULL
-                AND {excluded}
                 AND catalog.last_check = checks.id
-                AND checks.created_at <= $1
+                AND checks.next_check_at >= $1
```
maudetes, line 78: It's actually nice that this logic is kept simple with the next_check_at strategy! :clap:

maudetes (approved):
Thank you for this PR! I like the logic!

I think we may want to test this PR in our dev env and monitor the `next_check_at` values? Since it's quite difficult to test locally actually

Maybe we want to add some tests!

udata_hydra/analysis/resource.py
bolinocroustibat: I thought about the same, but thought also we should do that later. There are several places in the code where we could avoid re-requesting the check from the DB, but instead passing the record (or the dict) from one method to another.
Let's optimize that later indeed.

bolinocroustibat:
> Thank you for this PR! I like the logic!
> 
> I think we may want to test this PR in our dev env and monitor the `next_check_at` values? Since it's quite difficult to test locally actually
> 
> Maybe we want to add some tests!

Thanks for the review. I wanted to confirm the logic first, now currently fixing/adding tests - in the same PR.

PR #165, opened 2024-09-12
fix: use ENVIRONMENT from config file instead of env var

bolinocroustibat:
Related to an infrastruture change/PR.


magopian (approved):
Not sure I can comment much on this, nothing seems to be out of place, but I have no idea what this change does, nor if it's working properly ;)

PR #167, opened 2024-09-17
Add checks aggregate route

maudetes:
It allows to group_by on a column and count the check occurrences for a selected date.
Useful for debugging current checks trends, ex:

* http://localhost:8000/api/checks/aggregate?group_by=domain&created_at=today
* http://localhost:8000/api/checks/aggregate?group_by=status&created_at=today
* http://localhost:8000/api/checks/aggregate?group_by=headers-%3E%3E%27content-type%27&created_at=2024-09-05

## Notes
* This is probably not the most REST logic route, so I would be interested in alternative implementations. 
* Param parsing is very naive currently and would fail on non valid values
* I was wondering if we should follow these group_by indicators in a Munin plugin instead?


README.md
```diff
@@ -116,6 +116,7 @@ The API serves the following endpoints:
 *Related to checks:*
 - `GET` on `/api/checks/latest/?url={url}&resource_id={resource_id}` to get the latest check for a given URL and/or `resource_id`
 - `GET` on `/api/checks/all/?url={url}&resource_id={resource_id}` to get all checks for a given URL and/or `resource_id`
+- `GET` on `/api/checks/aggregate/?group_by={column}&created_at={date}` to get checks occurences grouped by a `column` for a specific `date`
```
bolinocroustibat, line 119: Not really related to this PR, but realizing now it would be better to write the URLs with GET parameters without trailing slashes in the documentation (like `/api/checks/aggregate?group_by={column}&created_at={date}`) since now the API supports both, and since GET parameters generally don't follow any trailing slashes.
Can make another PR to take care of that later.

bolinocroustibat, line 119: What about `/api/checks?group_by={column}&created_at={date}` instead? Maybe it is as explicit?
Isn't the aggregation an internal mecanism which doesn't need to be explicited in the URL?
Or, do we need to have an `aggregate` in the URL to differentiate other future type of search parameters?
Just raising the question, but might be more difficult to separate the routes in the code that way.

bolinocroustibat, line 119: I guess {date} is supposed to be an time interval (since `today` is possible)? If yes, can it not only a day, but also an hour, a month or a year? Maybe it would be good to specify what's possible in terms of time format interval in the doc, and maybe even to have specific error responses in case of not accepted time interval.

bolinocroustibat (approved):
LGTM, great job! A few optional questions/remark if we want to go further.

README.md
maudetes: I wanted to start with this approach, but I had some considerations:
* the payload would vary greatly depending on `group_by` being used or not.
* It also does not actually return `checks` objects per se, it returns count aggregation.
* Since we did not have any existing `GET /api/checks`, it seemed weird to require a `group_by` parameter on this route. It made me think that `/api/checks/latest` and `/api/checks/all` should probably be `/api/checks` with additional relevant params (pagination would do actually). Thus going with `/aggregate` path for now. A refactoring may follow to standardize `/api/checks` routes maybe.

maudetes: :+1: Done in https://github.com/datagouv/hydra/pull/167/commits/4e81f31f43177531ed5ec99451a1a0c97165eb6f.

maudetes: My personal use case was `today` (or latest) but it seemed fair and easy to allow to pick another date.
It is currently not a datetime interval since it's used as `created_at::date = $1`.

Making it an interval would probably require a `start` and `end` param I guess. Maybe we could have a start at `now - 1 day` and end at `now` to replicate the today (or other latest heuristic) default?

bolinocroustibat: I don't think we need to make it an interval with `start` and `end` params yet, what I meant is `today`is technically an interval (from midnight to midnight). So what other similar "one-word" interval it accepts and doesn't, should be documented. Does that make sense?

bolinocroustibat: Fair enough then, let's keep it that way. Thanks for the explanation!

bolinocroustibat: Perfect, thanks! There is a little typo, just commented

bolinocroustibat:
By the way, it's much better than this:

https://x.com/excid3/status/1836042970472075673

README.md
maudetes: I am leaving it unresolved as it may be useful reference for future readers.

maudetes: Done in https://github.com/datagouv/hydra/pull/167/commits/2aba2ee4ed35889518a148ed3884faccfcbb64ba

Issue #145, opened 2024-08-27
hydra: Convert to a package-manager-agnostic, PEP-621 standard Python project metadata


bolinocroustibat:
Right now, the hydra project can only be installed using Poetry, without giving any choice of the package manager to the developer or infra.

Also, Poetry <2 doesn't use the [PEP-621 standard](https://peps.python.org/pep-0621/) to edit the `pyproject.toml` project file (Poetry was created before the standardization of `pyproject.toml` and Poetry >=2 doesn't exist _yet_), so the project is currently not a standard Python project.

We can refactor the dependencies and metadata files to use the Python standards and the CI file to use those files (so the CI would have not to use Poetry).

This would still use the modern `pyproject.toml` standard, and `requirements.txt` as a lock file (like NPM `package.json` and `package-lock.json`).

PR: https://github.com/datagouv/hydra/pull/143

PR #170, opened 2024-09-24
feat: PEP-621 standard Python project metadata (#143)

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/145

Since [Poetry](https://python-poetry.org) 2.0.0 which was release in January 2025, [Poetry can now manage standardized `pyproject.toml` files.](https://python-poetry.org/blog/announcing-poetry-2.0.0).
So this PR aims to use a standardized `pyproject.toml` that can be used by any package manager and not only Poetry, by:

- Refactor `pyproject.toml` to become a package-manager agnostic project file
- Update README and documentation 
- Add a install Poetry step in the CI, as the `cimg/python` image doesn't uses Poetry 2.0


.circleci/config.yml
```diff
@@ -33,6 +33,10 @@ jobs:
             - << pipeline.parameters.cache-prefix >>-{{ arch }}-{{ .Branch }}
             - << pipeline.parameters.cache-prefix >>-{{ arch }}-{{ .Environment.BASE_BRANCH }}
             - << pipeline.parameters.cache-prefix >>-{{ arch }}-
+      - run:
+          name: Install Poetry
+          command: |
+            curl -sSL https://install.python-poetry.org | python3 -
```
magopian, line 39: Is there a way to pin the version or something? Would we want to pin it, to avoid a failure in the future if a new major, breaking version comes out?

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Add indexes to improve resource filtering and batch selection [#240](https://github.com/datagouv/hydra/pull/240)
 - Parallelize tests in CI [#238](https://github.com/datagouv/hydra/pull/238)
 - Refactor analysis logic to remove 5 non necessary queries, using the existing data in the code instead of re-querying it [#227](https://github.com/datagouv/hydra/pull/227)
+- Use a package-manager-agnostic pyproject.toml instead of a Poetry pyproject.toml [#143](https://github.com/datagouv/hydra/pull/143)
```
magopian, line 20: Should that link be changed to point to the PR `170` instead of `143`?

pyproject.toml
```diff
@@ -69,14 +79,10 @@ markers = [
 addopts = "--cov=udata_hydra/ --cov-report term-missing"
 
 [tool.ruff]
-lint = { extend-select = ["I"], ignore = ["F401"] } # ["I"] is to also sort imports with an isort rule # TODO: remove ignore F401 later
+lint = { extend-select = ["I"] }
+# ["I"] is to also sort imports with an isort rule
```
magopian, line 83: Nit: should that be on the same line as the previous line that it's commenting, or at least on the line above?

magopian (approved):
Nice job!
A couple of nits, and a CI failure: `The fields ['authors', 'description', 'name', 'version'] are required in package mode.`
This is very good step in the right direction 💯

.circleci/config.yml
bolinocroustibat: Good point. Although there weren't any breaking change except betwen 1.x.x and 2.0.0 I think we should installe Poetry with the latest 2.0.x.

pyproject.toml
bolinocroustibat: Yes, indeed! Not sure it was re-formatted automatically like this.

pyproject.toml
```diff
@@ -59,6 +63,12 @@ module = [
 ]
 ignore_missing_imports = true
 
+
+[tool.poetry.scripts]
```
hacherix, line 67: We can use `[project.scripts]` as well now 😄

hacherix (approved):

pyproject.toml
bolinocroustibat: Indeed! thanks

bolinocroustibat:
I think this can be merged now, what do you think @hacherix @Pierlou @maudetes ?

hacherix (approved):

PR #171, opened 2024-09-26
refactor: refactor function to get no_backoff domains and add PostgreSQL indexes to improve DB queries perfs

bolinocroustibat:
This PR aims to drastically improve DB queries performance in two ways:
- Refactor function to get no_backoff domains, in order to avoid unnecessary queries
- Add PostgreSQL indexes to improve performance when querying:
    - add indexes on `checks` table for `resource_id` and `created_at` columns
    - add index on `catalog` table for `last_check` column

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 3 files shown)


maudetes (approved):
Thank you! :rocket:

PR #174, opened 2024-09-26
refactor: refactor purge_checks CLI to use a date limit instead of a number

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/172

### DIFF
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -252,19 +252,13 @@ async def migrate(skip_errors: bool = False, dbs: list[str] = ["main", "csv"]):
 
 
 @cli
-async def purge_checks(limit: int = 2):
-    """Delete past checks for each resource_id, keeping only `limit` number of checks"""
-    q = "SELECT resource_id FROM checks GROUP BY resource_id HAVING count(id) > $1"
+async def purge_checks(days_until: int = 60):
+    """Delete outdated checks that are more than `days_until` days old"""
     conn = await connection()
-    resources = await conn.fetch(q, limit)
-    for r in resources:
-        resource_id = r["resource_id"]
-        log.debug(f"Deleting outdated checks for resource {resource_id}")
-        q = """DELETE FROM checks WHERE id IN (
-            SELECT id FROM checks WHERE resource_id = $1 ORDER BY created_at DESC OFFSET $2
-        )
-        """
-        await conn.execute(q, resource_id, limit)
+    log.debug(f"Deleting checks that are more than {days_until} days old")
+    await conn.execute(
+        f"""DELETE FROM checks WHERE created_at < now() - interval '{days_until} days'"""
+    )
 
 
 @cli
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -7,6 +7,7 @@ from tempfile import NamedTemporaryFile
 
 import aiohttp
 import asyncpg
+from asyncpg import Record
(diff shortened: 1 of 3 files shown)


udata_hydra/cli.py
```diff
@@ -252,19 +252,13 @@ async def migrate(skip_errors: bool = False, dbs: list[str] = ["main", "csv"]):
...
-        log.debug(f"Deleting outdated checks for resource {resource_id}")
-        q = """DELETE FROM checks WHERE id IN (
-            SELECT id FROM checks WHERE resource_id = $1 ORDER BY created_at DESC OFFSET $2
-        )
-        """
-        await conn.execute(q, resource_id, limit)
+    log.debug(f"Deleting checks that are more than {days_until} days old")
+    await conn.execute(
+        f"""DELETE FROM checks WHERE created_at < now() - interval '{days_until} days'"""
+    )
```
magopian, line 261: Nit: do we want another log showing the number of deleted checks? Would that be useful? Maybe something like
```Deleting 12 outdated checks. 1253 remaining.```

**Review by magopian — APPROVED** [review]
Excellent work 👍 
🚢

File: udata_hydra/cli.py [hunk 2]
```diff
@@ -252,19 +252,13 @@ async def migrate(skip_errors: bool = False, dbs: list[str] = ["main", "csv"]):
 
 
 @cli
-async def purge_checks(limit: int = 2):
-    """Delete past checks for each resource_id, keeping only `limit` number of checks"""
-    q = "SELECT resource_id FROM checks GROUP BY resource_id HAVING count(id) > $1"
+async def purge_checks(days_until: int = 60):
```
maudetes, line 255: I'm not sure about `days_until` wording. How about `retention_days`? (It's the wording we're using for [harvest jobs retention policy](https://github.com/opendatateam/udata/blob/c7bd1c28c399552164ccd5d448805c76e8bfeaa7/udata/harvest/actions.py#L168))

udata_hydra/cli.py
maudetes: (If we get a return value it would be nice to print, I would not start another query to show the number remaining though)

maudetes (approved):
The query is way more straightforward this way, thank you!

udata_hydra/cli.py
bolinocroustibat: Yep, that sounds more clear. Let me update that

bolinocroustibat: Done

udata_hydra/cli.py
bolinocroustibat: I think it's a good idea and could be very helpful. Since it's a query which is not run very often, we can afford it, if it's in the same query.

bolinocroustibat: I just updated it. Let me know what you think

magopian (approved):
Nice! I found out that you may be able to get the number of affected rows straight away, not sure if it's applicable here? https://github.com/MagicStack/asyncpg/issues/311#issuecomment-395449587

bolinocroustibat:
> Nice! I found out that you may be able to get the number of affected rows straight away, not sure if it's applicable here? [MagicStack/asyncpg#311 (comment)](https://github.com/MagicStack/asyncpg/issues/311#issuecomment-395449587)

Yes I tried that first, but the status format was pretty non standard and annoying to parse. It seems from the conversation I'm not the only one!

Issue #180, opened 2024-09-30
hydra: issue with `analyse-csv` CLI


bolinocroustibat:
Error when running analysis-csv CLI:

```
$ udata-hydra analyse-csv --check-id=13853277
Traceback (most recent call last):
  File "asyncpg/protocol/prepared_stmt.pyx", line 175, in asyncpg.protocol.protocol.PreparedStatementState._encode_bind_msg
  File "asyncpg/protocol/codecs/base.pyx", line 227, in asyncpg.protocol.protocol.Codec.encode
  File "asyncpg/protocol/codecs/base.pyx", line 129, in asyncpg.protocol.protocol.Codec.encode_scalar
  File "asyncpg/pgproto/./codecs/int.pyx", line 54, in asyncpg.pgproto.pgproto.int4_encode
TypeError: an integer is required (got type str)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/srv/hydra-hydra/bin/udata-hydra", line 8, in <module>
    sys.exit(run())
  File "/srv/hydra-hydra/lib/python3.9/site-packages/minicli/__init__.py", line 161, in run
    command.func(command, **shared)
  File "/srv/hydra-hydra/lib/python3.9/site-packages/minicli/__init__.py", line 45, in invoke
    return self(*args, **kwargs)
  File "/srv/hydra-hydra/lib/python3.9/site-packages/minicli/__init__.py", line 29, in __call__
    asyncio.get_event_loop().run_until_complete(res)
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/srv/hydra-hydra/lib/python3.9/site-packages/udata_hydra/cli.py", line 161, in analyse_csv_cli
    await analyse_csv(check_id=check_id, url=url, debug_insert=debug_insert)
  File "/srv/hydra-hydra/lib/python3.9/site-packages/udata_hydra/analysis/csv.py", line 115, in analyse_csv
    check: dict = await Check.get(check_id) if check_id is not None else {}
  File "/srv/hydra-hydra/lib/python3.9/site-packages/udata_hydra/db/check.py", line 23, in get
    return await connection.fetchrow(q, check_id)
  File "/srv/hydra-hydra/lib/python3.9/site-packages/asyncpg/connection.py", line 749, in fetchrow
    data = await self._execute(
[...]

PR #181, opened 2024-09-30
fix: fix analysis csv CLI

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/180

- Fix typing issue using minicli for analysis-csv CLI
- Fix method to get record columns in analysis-csv method
- Secure some cases in analysis-csv logic
- Add basic tests for analyse-csv CLI

### DIFF
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -155,10 +155,11 @@ async def check_resource(resource_id: str, method: str = "get"):
 
 @cli(name="analyse-csv")
 async def analyse_csv_cli(
-    check_id: int | None = None, url: str | None = None, debug_insert: bool = False
+    check_id: str | None = None, url: str | None = None, debug_insert: bool = False
 ):
     """Trigger a csv analysis from a check_id or an url"""
-    await analyse_csv(check_id=check_id, url=url, debug_insert=debug_insert)
+    check_id = int(check_id) if check_id else None
+    await analyse_csv(check_id, url, debug_insert)
 
 
 @cli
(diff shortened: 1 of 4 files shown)


udata_hydra/cli.py
```diff
@@ -155,10 +155,11 @@ async def check_resource(resource_id: str, method: str = "get"):
 
 @cli(name="analyse-csv")
 async def analyse_csv_cli(
-    check_id: int | None = None, url: str | None = None, debug_insert: bool = False
+    check_id: str | None = None, url: str | None = None, debug_insert: bool = False
```
maudetes, line 158: Hmm is this a minicli issue? It wouldn't cast it to `int` due to `| None`?

maudetes (approved):
Thank you for the fix!

udata_hydra/cli.py
bolinocroustibat: No I don't think it was a minicli issue. But all arguments of minicli are strings by default, so it needs to be casted as int (which I did on the fist line of the method). Type hinting argument cannot cast or change the type, it was useless (and false).

maudetes: I think `size` in `csv_sample` command is used as int directly?

bolinocroustibat: Good point, my guess is the type hint is wrong, and it's indeed being manipulated in the code as a string.
We should check and fix that.

bolinocroustibat: OK, just double-checked, you are right, it's seems to be related to the `int | None`, which seems to changes minicli's casting behaviour. It seems to be a bug from minicli. We already had a very similar bug before. I will created a ticket.

bolinocroustibat: OK thanks for spotting, I just created an issue: https://github.com/pyrates/minicli/issues/15

In the meantime we can continue using hydra CLI as it is.

PR #182, opened 2024-09-30
feat: save git commit hash in CI and use it for health check

bolinocroustibat:
Save git commit hash in CI and use it for config and health check, in order to be able to check quickly which version is deployed without having to deal with comparing the dev versions, as an attempt to improving overall developer experience.

Note: not 100% sure the file is imported in the build :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,9 @@ jobs:
           command: |
             poetry config virtualenvs.in-project true
             poetry install
+      - run:
+          name: Save commit hash in a file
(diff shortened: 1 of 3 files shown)


magopian (approved):
That's a clever trick, well done!

🚢

bolinocroustibat:
> That's a clever trick, well done!
> 
> 🚢

Too bad, it actually doesn't work: https://dev-crawler.data.gouv.fr/api/health/

Not sure if it's imported in the build?

PR #183, opened 2024-10-01
tests: coverage fails if coverage falls under 88%

bolinocroustibat:
Coverage fails if coverage falls under 88%, which is the current tests coverage.

Not exactly sure if this would make the CI tests fail, but it's a good security feature for next development as it will at least show clearly in the CLI if coverage has lowered since that commit.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "catalog_harvested: use catalog_harvested.csv as source",
 ]
-addopts = "--cov=udata_hydra/ --cov-report term-missing"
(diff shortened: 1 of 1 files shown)


magopian (approved):
I like this ;)
It might make it difficult in the future when modifying somewhat small pieces of code, but I believe it's a least worth giving it a try ;)

🚢

maudetes:
> It might make it difficult in the future when modifying somewhat small pieces of code, but I believe it's a least worth giving it a try ;)

I think that we can decrease the coverage requirement if needed but at least it will be a conscious action.

bolinocroustibat:
> > It might make it difficult in the future when modifying somewhat small pieces of code, but I believe it's a least worth giving it a try ;)
> 
> I think that we can decrease the coverage requirement if needed but at least it will be a conscious action.

Yes, exactly.

PR #184, opened 2024-10-01
feat: add a quiet option for purge_check and purge_csv_table CLIs

bolinocroustibat:
Similar to https://github.com/datagouv/data.gouv.fr/issues/1354 which was closed by https://github.com/datagouv/hydra/pull/95:

Add a `quiet` arg for `purge_check` and `purge_csv_table` CLIs, which makes the setting log level on `ERROR`, so that the CLI doesn't output logs by default and thus the infra doesn't send an email on each run of the CLI, unless there is an error.


maudetes (approved):
Thank you :pray:

PR #185, opened 2024-10-02
fix: fix commit hash in health check

bolinocroustibat:
- Fix [PR #182](https://github.com/datagouv/hydra/pull/182) by putting the shortened commit hash in the [local part of the version number](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers) (e.g. `2.0.1.dev4451+f6a9b94`), as brilliantly suggested by @maudetes

- Update changelog, which has not been updated in [PR #182](https://github.com/datagouv/hydra/pull/182)


maudetes:
As a suggestion, we could also decide to add the commit sha in build version in CI instead: https://github.com/datagouv/hydra/blob/396de866abd0d47408aa572a05c6c92e19c05027/.circleci/config.yml#L99

bolinocroustibat:
> As a suggestion, we could also decide to add the commit sha in build version in CI instead:
> 
> https://github.com/datagouv/hydra/blob/396de866abd0d47408aa572a05c6c92e19c05027/.circleci/config.yml#L99

That's a good idea, but if the commit hash is added at the end of the version (like  or `2.0.1.dev4451.381a85119c5b77fada04dcf543ebaf9d98d880ef` or even `2.0.1.dev381a85119c5b77fada04dcf543ebaf9d98d880ef'`, as I don't see any other possibilities for now) that would lead to a [non-PEP440-compliant version number](https://packaging.python.org/en/latest/specifications/version-specifiers/), since commit hash is not an integer but an alphanumerical string.

Invalid PEP 440 version: '2.0.1.dev381a85119c5b77fada04dcf543ebaf9d98d880ef'

maudetes:
From what I understand, I think you can add a [local part](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers), with `+`. Ex  `2.0.1.dev4451+f6a9b9e`
Not sure it would be cleaner to manipulate though.

I think you're creating .git-commit at the root of `hydra` repo and not inside `udata_hydra` package (you can actually open the .whl archive and see what's inside).

bolinocroustibat:
> From what I understand, I think you can add a [local part](https://packaging.python.org/en/latest/specifications/version-specifiers/#local-version-identifiers), with `+`. Ex `2.0.1.dev4451+f6a9b9e` Not sure it would be cleaner to manipulate though.

Ah, well spotted! That seems perfect for our case. Thank you!
Here is the PR updated with that idea.

maudetes (approved):
Thank you! :pray:

magopian (approved):
LGTM

PR #186, opened 2024-10-02
ci: trigger GitLab deployment pipeline when pushing

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1465.
Goes hand-to-hand with the GitLab MR.

For now, the trigger is sent only when we push on `main`.
We can imagine other policies?

**EDIT:**

Refactored entirely to use the GitLab _scaffolding_ deploy script repository instead, as the intermediary between this CI and GitLab API for infra repo CI.

Remarks:
- As a first version, this only triggers a call to the deploy script (and thus to GitLab for deploying on dev environment) when pushing on main. Later, we can imagine deploying easily, I suggest to do it in another PR.
- This PR also includes a better separation of the build job into smaller steps, clearer and easier to debug. This cleaning job can be more clearly done in this dedicated [PR #238](https://github.com/datagouv/hydra/pull/238), so if easier to review we can review and merge [#238](https://github.com/datagouv/hydra/pull/238) before the present one 
- Once merged, we will do the same thing on repo udata and any other suggested
- This has been partially tested, it successfully triggers the simple-scaffold script: https://app.circleci.com/pipelines/github/datagouv/hydra/1765/workflows/d04e28c0-7bf3-481f-af57-c0f660be8ab0/jobs/6328. It needs to be rested with the full workflow, once the GitLab API integration in CI is ready on infra's side.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,7 +137,8 @@ jobs:
           name: Configure the SSH deploy private key
           command: |
             mkdir -p ~/.ssh
-            echo "$DEPLOY_PRIVATE_KEY" > ~/.ssh/id_ed25519
+            # CircleCI doesn't accept multiple lines in a single environment variable, so the multiline private key must be base64 encoded, and then decoded here
+            echo "$DEPLOY_PRIVATE_KEY" | base64 -d > ~/.ssh/id_ed25519
             chmod 600 ~/.ssh/id_ed25519
             ssh-keyscan -t rsa gitlab.com >> ~/.ssh/known_hosts
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,17 +107,11 @@ jobs:
             echo "Commit hash: ${CIRCLE_SHA1:0:7}"
             echo "Tag: $CIRCLE_TAG"
       - run:
-          name: Build a distributable package
+          name: Build a distributable package with Poetry
           command: |
             RELEASE_VERSION=$(cat version.txt)
-            if [[ $CIRCLE_TAG ]]; then
-                # This is a tagged release, version has been handled upstream
-                poetry build
-            else
-                # Relies on a dev version like "1.2.1.dev" by default
-                poetry version $RELEASE_VERSION
-                poetry build
-            fi
+            poetry version $RELEASE_VERSION
+            poetry build
       - store_artifacts:
           path: dist
       - persist_to_workspace:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,6 +164,12 @@ jobs:
           command: |
             RELEASE_VERSION=$(cat version.txt)
             cd scaffold
+            # Run the script that triggers the Gitlab CI/CD pipeline.
+            # The script args are, in order:
+            # - hydra: the name of the project to deploy (APP_NAME)
+            # - $RELEASE_VERSION: the version to deploy (RELEASE_VERSION)
+            # - env: the environment to deploy to (ENV)
+            # - "": the deploy variables (VARS)
             ./scripts/gitlab-ci-pipeline.sh hydra $RELEASE_VERSION dev ""
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -137,6 +137,7 @@ jobs:
           name: Configure the SSH deploy private key
           command: |
             mkdir -p ~/.ssh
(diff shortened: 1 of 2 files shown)


maudetes:
I'd be willing to speak about the process live! I think having the job run after the `publish` would allow us to deploy the right pypi .dev version :)

magopian (approved):
Looks nice! Does it work? Is there any way to test it before it's merged in prod?

bolinocroustibat:
> I'd be willing to speak about the process live! I think having the job run after the `publish` would allow us to deploy the right pypi .dev version :)

Makes sense. Just updated so that the "deploy" job runs after the "publish" job and depends on it. LMKWYT!

.circleci/config.yml
```diff
@@ -126,6 +137,35 @@ jobs:
...
+      - run:
+          name: Clone scaffold repository
+          command: |
+            git clone --quiet --depth 1 $SCAFFOLD_REPO_SSH_URL scaffold
+      - run:
+          name: Trigger Gitlab CI/CD pipeline for Hydra to deploy to dev environment
+          command: |
+            RELEASE_VERSION=$(cat version.txt)
+            cd scaffold
+            ./scripts/gitlab-ci-pipeline.sh hydra $RELEASE_VERSION dev ""
```
maudetes, line 167: Maybe add an explicit comment about the expected params? I can't easily know what `""` stands for.

.circleci/config.yml
```diff
@@ -85,27 +85,37 @@ jobs:
...
-                export VERSION=$(poetry version -s)$CIRCLE_BUILD_NUM+${CIRCLE_SHA1:0:7}
+                export RELEASE_VERSION=$(poetry version -s)$CIRCLE_BUILD_NUM+${CIRCLE_SHA1:0:7}
             fi
-            # Display some debug info
-            echo "Building a wheel release with version $VERSION, build number: $CIRCLE_BUILD_NUM, commit hash: ${CIRCLE_SHA1:0:7}, tag: $CIRCLE_TAG."
-            # Build a wheel release
+            # Save version to a file that will be persisted
+            echo "$RELEASE_VERSION" > version.txt
+      - run:
+          name: Display build info for debugging
```
maudetes, line 102: :+1: It may prove quite useful for future debug!

.circleci/config.yml
```diff
@@ -85,27 +85,37 @@ jobs:
...
+          name: Build a distributable package
+          command: |
+            RELEASE_VERSION=$(cat version.txt)
             if [[ $CIRCLE_TAG ]]; then
                 # This is a tagged release, version has been handled upstream
                 poetry build
             else
                 # Relies on a dev version like "1.2.1.dev" by default
-                poetry version $VERSION
+                poetry version $RELEASE_VERSION
```
maudetes, line 118: Shouldn't we use `poetry version $RELEASE_VERSION` whether we have a `CIRCLE_TAG` or not?
It would allow for a deduplicated code et we would make sure that when we trigger `gitlab-ci-pipeline.sh`, it uses the `RELEASE_VERSION` that is always the one that's being built.

maudetes (commented):
:rocket: I will let @jordanguedj approve but it seems readable to me! Great job!

.circleci/config.yml
bolinocroustibat: Indeed!

jordanguedj (approved):

.circleci/config.yml
```diff
@@ -126,8 +134,48 @@ jobs:
...
+          command: |
+            RELEASE_VERSION=$(cat version.txt)
+            cd scaffold
+            # Run the script that triggers the Gitlab CI/CD pipeline.
+            # The script args are, in order:
+            # - hydra: the name of the project to deploy (APP_NAME)
+            # - $RELEASE_VERSION: the version to deploy (RELEASE_VERSION)
+            # - env: the environment to deploy to (ENV)
+            # - "": the deploy variables (VARS)
+            ./scripts/gitlab-ci-pipeline.sh hydra $RELEASE_VERSION dev ""
```
jordanguedj, line 175: I think `APP_NAME` should not be hard-coded but at least configured via a Github variable, as well as the environment (even if we want to enable it only for `dev`, we should be able to enable it for other environments in the future).

jordanguedj (changes requested):
Thanks for this!
However, even if we don't want to run this pipeline for another environment than `dev` (for now), this version of `.circleci/config.yml` will only and always apply to `dev` (`dev` environment shouldn't be hard-coded). IMO, we should have something dynamic and ready to handle other environments.

PR #187, opened 2024-10-03
fix: fix wrong resource status

bolinocroustibat:
Fix Sentry issue https://errors.data.gouv.fr/organizations/sentry/issues/145139/:

- A resource status was not the right one among valid resource statuses
- One resource status was not being used


Pierlou (approved):

maudetes:
Manually applied a migration in prod env for resources stuck with this erroneous status:
```
update catalog set status = null where status = 'TO_PROCESS_RESOURCE';
UPDATE 28258
```

PR #188, opened 2024-10-03
clean: more informative error relative to check resource CLI

bolinocroustibat:
Relative to Sentry error https://errors.data.gouv.fr/organizations/sentry/issues/144725


Pierlou (approved):

maudetes:
Hmm shouldn't we have the stacktrace as well as variables on the error by the way?

PR #191, opened 2024-10-04
feat: add comment column/field to ressources exceptions

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/190.


udata_hydra/db/resource_exception.py
```diff
@@ -32,7 +32,9 @@ async def get_by_resource_id(cls, resource_id: str) -> Record | None:
             return await connection.fetchrow(q, resource_id)
 
     @classmethod
-    async def insert(cls, resource_id: str, table_indexes: dict[str, str] | None = {}) -> Record:
+    async def insert(
+        cls, resource_id: str, table_indexes: dict[str, str] | None = {}, comment: str | None = None
```
magopian, line 36: We don't want a [mutable default value](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) in the function definition, right? I know it was there previously, should we still fix it here as the function definition is updated?

udata_hydra/db/resource_exception.py
```diff
@@ -50,25 +52,20 @@ async def insert(cls, resource_id: str, table_indexes: dict[str, str] | None = {
...
-                return await connection.fetchrow(q)
+        async with pool.acquire() as connection:
+            q = """
+                INSERT INTO resources_exceptions (resource_id, table_indexes, comment)
+                VALUES ($1, $2, $3)
+                RETURNING *;
+            """
+            return await connection.fetchrow(
+                q, resource_id, json.dumps(table_indexes) if table_indexes else None, comment
+            )
```
magopian, line 63: Nice simplification 👍

udata_hydra/db/resource_exception.py
```diff
@@ -50,25 +52,20 @@ async def insert(cls, resource_id: str, table_indexes: dict[str, str] | None = {
...
 
     @classmethod
-    async def update(cls, resource_id: str, table_indexes: dict[str, str] | None = {}) -> Record:
+    async def update(
+        cls, resource_id: str, table_indexes: dict[str, str] | None = {}, comment: str | None = None
+    ) -> Record:
```
magopian, line 68: Same as above, do we want to take the opportunity to remove the default mutable value?

magopian (approved):
Very nice!
A couple of minor nits that you may want to address.

udata_hydra/db/resource_exception.py
bolinocroustibat: Good point, thank you! Fixed

PR #193, opened 2024-10-16
Add extra for parquet export

Pierlou:


bolinocroustibat (approved):
LGTM!!

udata_hydra/analysis/csv.py
```diff
@@ -99,7 +99,7 @@ async def notify_udata(check_id: int, table_name: str) -> None:
             },
         }
         if config.CSV_TO_PARQUET:
-            payload["parquet_id"] = table_name
+            payload["document"]['analysis:parquet_url'] = check["parquet_url"]
```
maudetes, line 102:
```suggestion
            payload["document"]['analysis:parsing:parquet_url'] = check["parquet_url"]
```

maudetes (approved):
Seems good!

We may add a test later on!

udata_hydra/migrations/main/20241016_add_parquet_columns.sql
```diff
@@ -0,0 +1,5 @@
+-- add parquet_url and parquet_size columns to resources/catalog table
+
+ALTER TABLE catalog
```
maudetes, line 3: I think `parquet` info should be stored in `checks` table as for `parsing_table` info

maudetes (approved):
Could we add some tests maybe?

udata_hydra/migrations/main/20241016_add_parquet_columns.sql
bolinocroustibat: Don't we want parquet files to be unique per resource, and replaced everytime there is a new analysis?

maudetes: I think having the information of the checks on which the parquet has been constructed is important.
If we want to move it, the logic should be the same as `parsing_table` in any case :)

bolinocroustibat: Done!

bolinocroustibat: For info this PR doesn't include the creation of a DB method helper to get the latest parquet url from a resource, I prefer having both PRs separated and it's not needed in the code yet. Let me know if it will be needed and I'll create a ticket about it

maudetes: :ok_hand: let's way for the need first indeed!

maudetes (approved):
Looks good to me! I've added https://github.com/datagouv/hydra/pull/193/commits/a11ee1100fb269e476a1c92658b0c0e4fbf52240 to expose the parquet url and size by API :)

PR #194, opened 2024-10-22
ci: fix circleci version on main branch

bolinocroustibat:


.circleci/config.yml
```diff
@@ -90,6 +90,10 @@ jobs:
             # Set the version
             if [[ $CIRCLE_TAG ]]; then
                 export VERSION=$CIRCLE_TAG
+            # for main branches, can't add the commit hash since it's not a valid format for publishing
+            elif [[ $CIRCLE_BRANCH == "main" ]]; then
```
maudetes, line 94: I think you can use `pipeline-branch` (defined [here](https://github.com/datagouv/hydra/blob/6d49cda6224eae3fb3405eb905abbdbd234c4e71/.circleci/config.yml#L11-L13)) instead of "main"

maudetes (approved):
Thank you for the fix!

.circleci/config.yml
bolinocroustibat: You mean `publish-branch` right? Yes good idea

maudetes: Yes indeed, my bad!

bolinocroustibat: Done

PR #195, opened 2024-10-22
fix: crawl resources with no last check

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/1542


maudetes (approved):
:pray: thank you for the quick fix! Could you add a test case that was failing before and is now fixed?

maudetes (approved):

PR #197, opened 2024-10-24
fix: fix empty table indexes

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/198:

- Don't put empty values in `tables_indexes` columns of `resources_exceptions` table in the DB inser/update requests (happening when POSTing/PUTing `resources_exceptions`)
- Add test case to test the value when it's set to None
- Add a few exceptions catching related to `resources_exceptions` on the go, to secure the code

Note: the production table has been already updated so that all `resources_exceptions` have "{}" as the empty value (same as the default SQL value) when empty, so the error is already fixed with the current `resources_exceptions`:
`UPDATE resources_exceptions SET table_indexes = '{}' WHERE table_indexes IS NULL;`

### DIFF
--- a/udata_hydra/analysis/csv.py
+++ b/udata_hydra/analysis/csv.py
@@ -130,13 +130,8 @@ async def analyse_csv(
     # If it is, get the table_indexes to use them later
     exception: Record | None = await ResourceException.get_by_resource_id(resource_id)
     table_indexes: dict | None = None
-    if exception:
-        try:
-            table_indexes: dict | None = json.loads(exception["table_indexes"])
-        except TypeError:
-            raise TypeError(
-                f"table_indexes should be a JSON object, got {exception.get('table_indexes')}"
-            )
+    if exception and exception.get("table_indexes"):
+        table_indexes = json.loads(exception["table_indexes"])
 
     timer = Timer("analyse-csv")
     assert any(_ is not None for _ in (check_id, url))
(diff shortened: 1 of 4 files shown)


udata_hydra/analysis/csv.py
```diff
@@ -129,7 +129,14 @@ async def analyse_csv(
...
     exception: Record | None = await ResourceException.get_by_resource_id(resource_id)
-    table_indexes: dict | None = json.loads(exception["table_indexes"]) if exception else None
+    table_indexes: dict | None = None
+    if exception:
+        try:
+            table_indexes: dict | None = json.loads(exception["table_indexes"])
+        except TypeError:
+            raise TypeError(
+                f"table_indexes should be a JSON object, got {exception.get('table_indexes')}"
+            )
```
maudetes, line 139: Could you do it like this?
```suggestion
            table_indexes: dict | None = json.loads(exception.get("table_indexes", "{}") or "{}")
```
exception["table_indexes"] is actually a *JSON string* and not a *JSON object*

maudetes (approved):
Thank you for the fix!

udata_hydra/analysis/csv.py
bolinocroustibat: I think we want the var table_indexes to actually be None if it's empty in the DB, so that it doesn't even go through the index creation logic, but we can do:

```python
if exception and exception.get("table_indexes"):
    print("something in table indexes")
    table_indexes = json.loads(exception.get("table_indexes", "{}"))
```

bolinocroustibat: Changed. Will merge in a bit if that's OK

PR #204, opened 2024-10-28
fix: fix type issues

bolinocroustibat:
Fix minor types issues.


Pierlou (approved):
Not sure why the inversion, but looks good to me!

bolinocroustibat:
> Not sure why the inversion, but looks good to me!
Argument order has been changed because only the last parameter is non mandatory

Issue #201, opened 2024-10-28
[hydra] Allow to crawl non-data.gouv URLs


Pierlou:
In local tests, it can be useful to trigger a crawl of URLs that are not of data.gouv resources (for instance a test file from demo). This used to be possible using CLI `poetry run analyse-csv --url <url>` before URLs/resource_ids where entangled, but now this raises [this error](https://github.com/datagouv/hydra/blob/ecbe288c115284026288ca60d6ff607815867548/udata_hydra/analysis/csv.py#L127) (which is expected).
Do we want to change this to allow crawling external URLs just to see how they are handled (csv-detection, parquet, postgres), maybe dropping all results eventually (check, inserted rows...)?

PR #205, opened 2024-10-28
Add option to force analysis even if resource has not changed

Pierlou:
We're missing this feature: forcing analysis on a resource, independently of whether its content has changed since last check.

### DIFF
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -137,7 +137,7 @@ async def crawl_url(url: str, method: str = "get"):
 
 
 @cli
-async def check_resource(resource_id: str, method: str = "get"):
+async def check_resource(resource_id: str, method: str = "get", force_analysis: bool = False):
     """Trigger a complete check for a given resource_id"""
     resource: asyncpg.Record | None = await Resource.get(resource_id)
     if not resource:
@@ -149,6 +149,7 @@ async def check_resource(resource_id: str, method: str = "get"):
             resource_id=resource_id,
             session=session,
             method=method,
+            force_analysis=force_analysis,
             worker_priority="high",
         )
 
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -137,7 +137,7 @@ async def crawl_url(url: str, method: str = "get"):
 
 
 @cli
-async def check_resource(resource_id: str, method: str = "get", force_analysis: bool = False):
+async def check_resource(resource_id: str, method: str = "get", force_analysis: bool = True):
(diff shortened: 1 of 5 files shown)


bolinocroustibat (approved):
LGTM. A few questions:
Is this related to the ability to manually check a URL even though the resource doens't exist in the DB? Should we also create a ticket about that to go with it (and it that case, it shouldn't update any resource status of course)

bolinocroustibat:
Would be nice to add a test about that, when resource has not changed but parameter `force_analysis` is passed as False, then True.

Pierlou:
> Is this related to the ability to manually check a URL even though the resource doens't exist in the DB? Should we also create a ticket about that to go with it (and it that case, it shouldn't update any resource status of course)

Nope, completely different topic. I have created [this issue](https://github.com/datagouv/hydra/issues/201), is that fine?

udata_hydra/cli.py
```diff
@@ -137,7 +137,7 @@ async def crawl_url(url: str, method: str = "get"):
 
 
 @cli
-async def check_resource(resource_id: str, method: str = "get"):
+async def check_resource(resource_id: str, method: str = "get", force_analysis: bool = False):
```
maudetes, line 140: I would even set `force_analysis` to True by default? I think if we're using a CLI to launch a check, I would say we probably want the full run

udata_hydra/crawl/check_resources.py
```diff
@@ -58,6 +58,7 @@ async def check_resource(
     url: str,
     resource_id: str,
     session,
+    force_analysis: bool = False,
```
maudetes, line 61: Could it be close to `worker_priority` in the params order since it's kind of the same vibe?

udata_hydra/routes/checks.py
```diff
@@ -64,6 +64,7 @@ async def create_check(request: web.Request) -> web.Response:
     try:
         payload: dict = await request.json()
         resource_id: str = payload["resource_id"]
+        force_analysis: bool = payload.get("force_analysis", False)
```
maudetes, line 67: Same here! Do we want it to be True by default?

maudetes (approved):
I have minor suggestions/questions :)

bolinocroustibat:
> > Is this related to the ability to manually check a URL even though the resource doens't exist in the DB? Should we also create a ticket about that to go with it (and it that case, it shouldn't update any resource status of course)
> 
> Nope, completely different topic. I have created [this issue](https://github.com/datagouv/hydra/issues/201), is that fine?

Definitely, thank you

PR #206, opened 2024-10-29
feat: statuses count in crawler status route

bolinocroustibat:


udata_hydra/routes/status.py
```diff
@@ -8,6 +8,27 @@
...
+        q = """
+            SELECT COALESCE(status, 'NULL') AS status, COUNT(*) AS count
+            FROM catalog
+            GROUP BY COALESCE(status, 'NULL');
+        """
+        rows = await connection.fetch(q)
+
+        for row in rows:
+            status = row["status"] if row["status"] != "NULL" else None
+            status_counts[status] = row["count"]
```
maudetes, line 27: Can't we do
```suggestion
    async with pool.acquire() as connection:
        q = """
            SELECT status AS status, COUNT(*) AS count
            FROM catalog
            GROUP BY status ;
        """
        rows = await connection.fetch(q)
        status_counts.update(dict(rows))
```
?

maudetes (approved):
Thank you! It can clearly be quite useful! :pray:

PR #208, opened 2024-10-30
clean: clean types and refactor some functions

bolinocroustibat:
Another cleaning PR:
- Remove useless types and import
- Add a few type hints
- Slight refactor of some functions

### DIFF
--- a/udata_hydra/routes/status.py
+++ b/udata_hydra/routes/status.py
@@ -8,27 +8,6 @@ from udata_hydra.db.resource import Resource
 from udata_hydra.worker import QUEUES
 
 
-async def get_status_counts(request: web.Request) -> dict[str, int]:
-    pool = request.app["pool"]
-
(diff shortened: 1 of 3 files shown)


udata_hydra/routes/status.py
```diff
@@ -60,14 +39,31 @@ async def get_crawler_status(request: web.Request) -> web.Response:
     rate_checked = round(stats_catalog["count_checked"] / total * 100, 1)
     rate_checked_fresh = round(count_checked / total * 100, 1)
 
+    async def get_resources_status_counts(request: web.Request) -> dict[str | None, int]:
```
magopian, line 42: Wow I'm so glad there's a type annotation here, I couldn't believe that we'd have `None` as a key, but we do, and it works, and it's what we want! 🤯

magopian (approved):
Nice cleanup! I like those <3

🚢

PR #211, opened 2024-10-30
fix: handle Sentry issue when check is not found

bolinocroustibat:
Handle Sentry issue [#148024](https://errors.data.gouv.fr/organizations/sentry/issues/148024/?project=21) with at least more explicit error reporting.

Should we just log a warning for that?

PR can be improved depending on what we decide.


maudetes (commented):
I am confused of when this would happen though

bolinocroustibat:
> I am confused of when this would happen though

Me too, but it seems it does happen, for some reason. Some race condition maybe?

PR #212, opened 2024-10-30
fix: handle Sentry issue for CSV downloading errors

bolinocroustibat:
Attempt to handle numerous Sentry errors similar to [#145755](https://errors.data.gouv.fr/organizations/sentry/issues/145755/?project=21&query=error.unhandled%3Atrue&referrer=issue-stream&statsPeriod=14d) in a more explicit and generic way:

- Catch the exception when the CSV URL is not reachable and rename it more clearly, so that it can be more instantaneously understood as non major when browsing Sentry
- New custom `IOException` using a new parent class `ExceptionWithSentryDetails` which sends details to Sentry as tags, also inherited by the recent custom `ParseException`
- Move the `errors.py` file from `/analysis` to `/utils`, since it's more broadly used throughout the code
- Use `sentry_sdk.Hub.current.client` instead of `config.SENTRY_DSN` to check if Sentry is correctly working, as it is the recommended way


udata_hydra/utils/file.py
```diff
@@ -52,15 +52,18 @@ async def download_resource(
...
+                async for chunk in response.content.iter_chunked(chunk_size):
+                    if max_size_allowed is None or i * chunk_size < max_size_allowed:
+                        tmp_file.write(chunk)
+                    else:
+                        tmp_file.close()
+                        log.warning(f"File {url} is too big, skipping")
+                        raise IOError("File too large to download")
+                    i += 1
+        except aiohttp.ClientResponseError as e:
+            raise IOError(f"Error downloading CSV: {e}")
```
maudetes, line 66: I don't think this makes the error more explicit than previously?
I think the issue is more about marking these as "expected" in sentry

maudetes (commented):
I think if we raise an error, we won't return the `tmp_file` and maybe won't remove it in the `finally` close?

udata_hydra/utils/file.py
bolinocroustibat: I think the error would be more explicit that way, if explained if the error title. For now in Sentry it's not easy to identify what is the issue is when listing the errors, so we can't see immediately it's not a major issue.

We could also create a specific exception which would include attributes to identify the ressource, as we did for the custom `ParseException`. What do you think?

bolinocroustibat: ...Updated!

bolinocroustibat:
> I think if we raise an error, we won't return the `tmp_file` and maybe won't remove it in the `finally` close?

Google point. Logic has been modified, try/catch block now has a finally block to close and return the temp file.

magopian (approved):
That should make it easier to grok through sentry errors, nice work!
🚢

bolinocroustibat:
> That should make it easier to grok through sentry errors, nice work! 🚢

@magopian Thanks! Should we do something similar for udata? What do you think? (I can do the PR if needed)

magopian:
I'm all up for it!

bolinocroustibat:
> I'm all up for it!

A proposal: https://github.com/opendatateam/udata/pull/3201

udata_hydra/utils/errors.py
```diff
@@ -4,54 +4,71 @@
...
+    pass
 
 
 async def handle_parse_exception(e: ParseException, table_name: str, check: Record | None) -> None:
     """Specific ParseException handling. Store error if in a check context. Also cleanup :table_name: if needed."""
     db = await context.pool("csv")
     await db.execute(f'DROP TABLE IF EXISTS "{table_name}"')
     if check:
-        if config.SENTRY_DSN:
+        if sentry_sdk.Hub.current.client:
```
maudetes, line 62: Why the change from `config.SENTRY_DSN` to `sentry_sdk.Hub.current.client`?

maudetes (approved):
Thank you for this additional sentry logging!

udata_hydra/utils/errors.py
bolinocroustibat: It seems to be the recommend way to test if the Sentry client is properly started, in several example like [that one](https://docs.sentry.io/platforms/python/configuration/draining/) and cover larger number of cases of Sentry not being properly initialized compared to `config.SENTRY_DSN`.
But I agree it weakens the readability. Should we revert that?

PR #213, opened 2024-10-30
fix: more explicit error reporting when sending to udata without raising errors for udata responding with a 404

bolinocroustibat:
Handle Sentry issue [#2333](https://errors.data.gouv.fr/organizations/sentry/issues/2333/?project=21) with at least more explicit error reporting.

Should we just log a warning for that?

PR can be improved depending on what we decide.


udata_hydra/utils/http.py
```diff
@@ -48,4 +48,11 @@ async def send(dataset_id: str, resource_id: str, document: dict) -> None:
     async with aiohttp.ClientSession() as session:
         async with session.put(uri, json=document, headers=headers) as resp:
             # we're raising since we should be in a worker thread
-            resp.raise_for_status()
+            if resp.status == 404:
+                raise IOError("Resource not found on udata")
+            elif resp.status == 502:
+                raise IOError("Resource has been deleted on udata")
```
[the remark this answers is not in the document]
bolinocroustibat: I think that was a mistake, this was to cover the 410 case. Well noted, fixed

bolinocroustibat:
> I think we could catch the 404 case and prevent raising but raise in all other cases?

That would be a good way to reduce Sentry errors. Changed

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Refactor ParseError to enrich Sentry with context and to inquire about Sentry errors like #4096  [#218](https://github.com/datagouv/hydra/pull/218)
 - Remove legacy routes [#203](https://github.com/datagouv/hydra/pull/203)
+- Handle Sentry issue #2333 with more explicit error reporting and not raising an error anymore for uadata 404 errors [#213](https://github.com/datagouv/hydra/pull/213)
```
maudetes, line 7: I think error #2333 will not exist anymore in the future, could you describe it shortly instead?
```suggestion
- Handle Sentry issue #2333 with more explicit error reporting and not raising an error anymore for udata 404 errors [#213](https://github.com/datagouv/hydra/pull/213)
```

maudetes (approved):

CHANGELOG.md
bolinocroustibat: Makes sense. Done!

PR #215, opened 2024-11-06
docs: update docstrings

bolinocroustibat:
Rewrite a few docstrings to use a more consistent syntax.

Since we use type hints extensively, we don't need to rewrite the types in the docstring, so we decided use the quite broadly standard [Google styleguide](https://google.github.io/styleguide/pyguide.html) for docstrings.


Pierlou (approved):
📖

PR #218, opened 2024-11-12
fix: refactor ParseError to enrich Sentry with context and to inquire about Sentry errors like #4096

bolinocroustibat:
- Refactor ParseError as a generic exception to enrich Sentry with context and to inquire about Sentry errors like [#4096](https://errors.data.gouv.fr/organizations/sentry/issues/4096) when analysing CSV and creating DB tables
- Update deprecated Sentry SDK method `set_extra()` to `set_tags()`
- Update deprecated Sentry SDK method `push_scope()` to `new_scope()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
+- Fix Sentry issue #4096 regarding duplicate columns names [#218](https://github.com/datagouv/hydra/pull/218)
 
 ## 2.0.5 (2024-11-08)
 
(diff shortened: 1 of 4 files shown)


Pierlou:
I am not sure how duplicate column names can happen here:  `csv-detective`  opens the file using `pandas` which has its own way of dealing with duplicate labels (adds `.X`, then `.X+1` to the duplicated name; BTW we could (should?) change this, because `tabular_api` doesn't like dots in column names). Plus `inspection["columns"]` is a `dict`, so keys can't be duplicated, therefore there should be no advantage of storing the keys in `seen`, they should appear only once right?
I'd like to reproduce the issue locally to see if it could be a fix in `csv-detective`, do you know which file lead to this (I didn't manage to find it in the issue)?

Pierlou (approved):
🚀

PR #221, opened 2024-11-18
chore: update python version from 3.11.9 to 3.11.10

bolinocroustibat:


.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ version: 2.1
 parameters:
   python-version:
     type: string
-    default: "3.11.9"
+    default: "3.11.10"
```
maudetes, line 7: I think the CI version should match the one in our production env :)

maudetes (approved):

.circleci/config.yml
bolinocroustibat: Good point. It's 3.11.2 in the production env, so we could downgrade the CI Python version to 3.11.2.

Although minor Python versions are fixes and are not supposed to break compatibility though, so I'm still hesitant.

WDYT?

PR #228, opened 2024-11-28
CLI to insert resource into local catalog

Pierlou:
This proved useful locally for me, to insert a new resource into my local catalog (or update an existing one), instead of having to do database shenaningans


bolinocroustibat (approved):
Quite useful indeed, thanks!
LGTM, but should we write a test for this new command in `tests/test_cli.py`?

Pierlou:
I don't like the `"default"` value in the CLI function, but it seemed to be the only way to make the test pass. We will fix this when https://github.com/datagouv/hydra/issues/232 is treated

PR #230, opened 2024-12-04
fix: refactor purge_csv_tables to fix dealdocks errors when purging CSV tables using atomic transactions

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/229.


udata_hydra/cli.py
```diff
@@ -291,16 +291,32 @@ async def purge_csv_tables(quiet: bool = False):
...
-    if len(tables_to_delete):
-        log.info(f"Deleted {len(tables_to_delete)} table(s).")
-    else:
+        try:
+            async with conn.transaction():
+                log.debug(f'Deleting table "{table}"')
+                await conn.execute(f'DROP TABLE IF EXISTS "{table}"')
+                await conn.execute("DELETE FROM tables_index WHERE parsing_table = $1", table)
+                await conn.execute(
+                    "UPDATE checks SET parsing_table = NULL WHERE parsing_table = $1", table
```
maudetes, line 307: Could we add a foreign key constraints as was done in https://github.com/datagouv/hydra/pull/195/, but to a table name?

maudetes (approved):
Thank you for the improved error management!

udata_hydra/cli.py
bolinocroustibat: Good idea, I suggest we do that later in another PR, after stabilising the current hydra and deploying the new features, though.
Here is the dedicated issue: https://github.com/datagouv/hydra/issues/231

jordanguedj (approved):

magopian (approved):
Nice work! I like the sprinkling of types here and there when you modify code, good habit!

PR #233, opened 2024-12-09
Remove minicli and use Typer

bolinocroustibat:


tests/test_cli/test_catalog.py
```diff
@@ -1,11 +1,13 @@
...
 import pytest
-from minicli import run
+from typer.testing import CliRunner
 
 from tests.conftest import RESOURCE_ID
+from udata_hydra.cli import insert_resource_into_catalog, load_catalog
 from udata_hydra.db.resource import Resource
 
 pytestmark = pytest.mark.asyncio
+runner = CliRunner()
```
Pierlou: This doesn't seem to be used in the code, is it called under the hood?

tests/test_cli/test_purge.py
```diff
@@ -168,6 +189,58 @@ async def test_purge_csv_tables_url_not_in_catalog(setup_catalog, db, fake_check
     assert res is None
 
 
+async def test_load_catalog_url_has_changed(setup_catalog, rmock, db, catalog_content):
```
Pierlou: These additions are because we didn't test these cases before? Maybe they would rather belong in `test_catalog.py`?

Pierlou (approved):
Thanks a lot for taking the time to clean up 🧹

Pierlou (approved):
✨

PR #236, opened 2024-12-18
Use pre-cast dataframe from csv-detective

Pierlou:
I wonder what is the best pratice for the requirements: should we specify `pandas` version or rely on the one installed by `csv-detective`

NB: ⚠️ do not merge before a real `csv-detective` version is released and put in the pyproject


pyproject.toml
```diff
@@ -12,23 +12,21 @@ aiocontextvars = "^0.2.2"
...
+csv-detective = "0.7.5.dev1113"
 dateparser = "^1.1.7"
 humanfriendly = "^10.0"
 marshmallow = "^3.14.1"
 minicli = "^0.5.3"
 minio = "7.2.7"
-pyarrow = "16.1.0"
-python-dateutil = "^2.8.2"
+pyarrow = "18.1.0"
+pandas = "2.2.0"
```
bolinocroustibat, line 22: I would rely on pandas installed by csv-detective, so we avoid future potential pandas versions conflicts (which are actually very common and annoying with pandas/numpy). I don't know which is the best practice but it would save us somes future headaches for sure. And since the features using pandas in the code are very related to csv-detective output, it also makes sense.

Pierlou: Makes sense, I'll change that (also the requirements of `csv-detective` are rather bad, I'll make a PR to improve them)

bolinocroustibat:
LGTM, OK to merge!

bolinocroustibat (approved):
Rocket Thumb

PR #238, opened 2024-12-20
ci: parellelize tests in CI

bolinocroustibat:
- Add install step cache fallback

- Parallelize the execution of tests among 4 CircleCI executors. This results in a gain of an average approx. ~40% on the total CI execution time (around 60% specifically on the tests job)
    - Note 1: For this, we need to deactivate the maximum coverage rule for test success, as each executor reach a different coverage. The gain makes it worth to deactivate this feature, as anyway we can monitor the coverage, and we can activate the minimum coverage threshold when running locally if necessary.
    - Note 2: it has also been tested with 2 executors, 3 executors and 10 executors, but 4 executors seems like a good balance for time gains

Before:
Screenshot 2024-12-20 at 20 32 49

After:
Screenshot 2024-12-20 at 20 32 38

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,15 +82,13 @@ jobs:
             # Find and split tests
             TESTFILES=$(find tests -name "test_*.py" | circleci tests split --split-by=timings)
             # Debug: Show what's in TESTFILES
-            echo "Files assigned to this executor:"
+            echo "Test files assigned to the executor ${CIRCLE_NODE_INDEX}/${CIRCLE_NODE_TOTAL}:"
             echo "$TESTFILES"
             # Run the found tests
             if [ -n "$TESTFILES" ]; then
-              echo "Running tests: $TESTFILES"
               poetry run pytest --junitxml=reports/python/tests.xml -p no:sugar --color=yes $TESTFILES
             else
               echo "No tests to run in this split"
-              find tests -name "test_*.py" | wc -l
               exit 1
             fi
       - store_test_results:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,9 +85,8 @@ jobs:
       - attach_workspace:
           at: .
       - run:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -77,7 +79,20 @@ jobs:
...
+            # Debug: Show what's in TESTFILES
+            echo "Files assigned to this executor:"
+            echo "$TESTFILES"
+            # Run the found tests
+            if [ -n "$TESTFILES" ]; then
+              echo "Running tests: $TESTFILES"
+              poetry run pytest --junitxml=reports/python/tests.xml -p no:sugar --color=yes $TESTFILES
+            else
+              echo "No tests to run in this split"
+              find tests -name "test_*.py" | wc -l
```
magopian, line 93: Is this a leftover debug?

magopian (approved):
Nice speedup, well done!

I guess the appropriate number of threads depends on the number of CPUs/cores that are available in a circleCI run?

Would using pytest-xdist make sense to "automate" the parallelization of tests instead of dealing with the executors "manually"?

.circleci/config.yml
bolinocroustibat: Well spotted, thanks. Fixed

bolinocroustibat:
> I guess the appropriate number of threads depends on the number of CPUs/cores that are available in a circleCI run?

No, I think each CircleCI executor is a separate execution environment, either a Docker container or a virtual machine. (https://circleci.com/docs/executor-intro/)
I have trouble to find documentation about the maximum number of executors we can use in free tier, but it seems to be at least 10 since I've been trying with 10 with no issue.

> Would using pytest-xdist make sense to "automate" the parallelization of tests instead of dealing with the executors "manually"?

Yes, that would also good so that it would run in parallel also locally, but that would mean some tests and code refactoring. This way is easier for now. Also I wonder if CircleCI executors have several CPUs/cores?

PR #243, opened 2025-01-29
fix: CI deployment versioning

hacherix:
Fix the `$RELEASE_VERSION` set in the `Build a distributable package as a wheel release with Poetry` CI job.


maudetes (approved):
Thank you for the quickfix!

Issue #242, opened 2025-01-20
CLI `check_resource` needs fixing


Pierlou:
https://github.com/datagouv/hydra/blob/833536dfc6b0d6f4c11ac119658583311fc7cc38/udata_hydra/cli.py#L141 doesn't work anymore because https://github.com/datagouv/hydra/blob/833536dfc6b0d6f4c11ac119658583311fc7cc38/udata_hydra/crawl/check_resources.py#L56 now takes a `Record` instead of a simple `resource_id` as an argument

PR #248, opened 2025-02-27
Fix various issues

Pierlou:
- https://github.com/datagouv/hydra/issues/242
- https://github.com/datagouv/hydra/issues/250

### DIFF
--- a/udata_hydra/cli.py
+++ b/udata_hydra/cli.py
@@ -179,14 +179,11 @@ async def analyse_csv_cli(
     if not check:
         if check_id:
             log.error("Could not retrieve the specified check")
-            return
         elif url:
             log.error("Could not find a check linked to the specified URL")
-            return
         elif resource_id:
             log.error("Could not find a check linked to the specified resource ID")
-            return
-
+        return
     await analyse_csv(check=check, debug_insert=debug_insert)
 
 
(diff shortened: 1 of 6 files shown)


udata_hydra/cli.py
```diff
@@ -157,24 +157,34 @@ async def check_resource(resource_id: str, method: str = "get", force_analysis:
...
         if checks and len(checks) > 1:
             log.warning(f"Multiple checks found for URL {url}, using the latest one")
         check = checks[0] if checks else None
+    if not check and resource_id:
+        check: Record | None = await Check.get_by_resource_id(resource_id)
     if not check:
-        log.error("No check found or URL provided")
+        if check_id:
+            log.error("Could not retrieve the specified check")
+        elif url:
```
hacherix, line 182: nit: Maybe leave them as `if` in case the 3 of them failed? And print the `check_id`, `url` or `resource_id` in the relevant log?

udata_hydra/cli.py
```diff
@@ -147,7 +147,7 @@ async def check_resource(resource_id: str, method: str = "get", force_analysis:
     async with aiohttp.ClientSession(timeout=None) as session:
         await crawl_check_resource(
             url=resource["url"],
-            resource_id=resource_id,
+            resource=resource,
```
hacherix, line 150: Nice catch! How this was even working until now?!

hacherix (approved):

udata_hydra/cli.py
Pierlou: The CLI commands are not used in production, we only use them locally (to test current developments) or on the prod server to check stuff in real conditions (actually that is something we're trying to avoid now, because most useful commands are accessible through the API, including a force analysis). But it's quite nice to have this available when you're developping!

PR #251, opened 2025-03-13
Reset stuck statuses

Pierlou:
I am wondering if we should also clean statuses that are not within the [current list](https://github.com/datagouv/hydra/blob/df2a916b2c77e595b40d50d2aa53924c580aa3ac/udata_hydra/db/resource.py#L9) in `Resource.clean_up_statuses`

### DIFF
--- a/udata_hydra/crawl/__init__.py
+++ b/udata_hydra/crawl/__init__.py
@@ -5,7 +5,6 @@ from asyncpg import Record
 from udata_hydra import config, context
 from udata_hydra.crawl.check_resources import check_batch_resources
 from udata_hydra.crawl.select_batch import select_batch_resources_to_check
-from udata_hydra.db.resource import Resource
 from udata_hydra.logger import setup_logging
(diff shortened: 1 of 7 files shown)


udata_hydra/crawl/__init__.py
```diff
@@ -25,6 +26,7 @@ async def start_checks(iterations: int = -1) -> None:
         )
 
         while iterations != 0:
+            await Resource.clean_up_statuses()
```
maudetes, line 29: Don't we want to run this in a job instead of running it at each batch?

maudetes (approved):
Looks nice, thank you! :pray:

udata_hydra/crawl/__init__.py
Pierlou: I don't know what's better, nor where it should go if we want it as a job 😅

maudetes (approved):
Thank you! Looks nice :rocket:

PR #252, opened 2025-03-14
Update harvest_modified_at on load catalog

maudetes:
For example, [this resource](https://www.data.gouv.fr/fr/datasets/liste-des-experiences-partagees-par-les-usagers/#/resources/91ea5a74-35f9-48d3-86cb-4407bf766927) stills has `analysis:last-modified-detection : harvest-resource-metadata` even though the dataset is not harvested anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Current (in progress)
 
-- Nothing yet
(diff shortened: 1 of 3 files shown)


Pierlou (approved):
Nice catch, thanks 🚜

PR #255, opened 2025-03-24
Prioritize new resources

Pierlou:


maudetes (commented):
Thank you for this PR!
I think we can try and got this direction :) It just depends how much jobs we would have in `high`, since currently only `send` is  high by default. Since now we would add some longer jobs in this queue, it may not always be at 0 all of the time.

I think we may want to forward the priority level in [`analyse_csv` here as well](https://github.com/datagouv/hydra/blob/f947e7e00990a1be0585d2fa89e2d098b0166e6c/udata_hydra/analysis/resource.py#L119)?

udata_hydra/analysis/resource.py
```diff
@@ -116,7 +116,13 @@ async def analyse_resource(
             # Change status to TO_ANALYSE_CSV
             await Resource.update(resource_id, data={"status": "TO_ANALYSE_CSV"})
             # Analyse CSV and create a table in the CSV database
-            queue.enqueue(analyse_csv, check=check, file_path=tmp_file.name, _priority="default")
+            is_priority = await Resource.get(resource_id, "priority")
```
maudetes, line 119: I think that at this point in time the resource priority as been removed? Or I guess it should since crawling has already occur.

Pierlou: You're very right, I missed the part where we reset the priority. I'll pass the priority downstream as an additionnal argument 🙏

udata_hydra/analysis/resource.py
```diff
@@ -116,7 +119,12 @@ async def analyse_resource(
             # Change status to TO_ANALYSE_CSV
             await Resource.update(resource_id, data={"status": "TO_ANALYSE_CSV"})
             # Analyse CSV and create a table in the CSV database
-            queue.enqueue(analyse_csv, check=check, file_path=tmp_file.name, _priority="default")
+            queue.enqueue(
+                analyse_csv,
+                check=check,
+                file_path=tmp_file.name,
+                _priority="high" if worker_priority == "high" else "default",
```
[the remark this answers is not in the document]
Pierlou: I don't know what it would imply exactly, since it would mean "low" priority for all csv analyses that are not new in catalog right?

maudetes (approved):
We can keep the worker priority as `default` or `low`, as you prefer

Issue #137, opened 2024-08-21
hydra: Ensure analysis payload is set to None when facing an analysis error


maudetes:
When we encounter a `"File too large to download"` on [analysis](https://github.com/datagouv/hydra/blob/f7b1fe7f23bdd1a2d481e67e4a2249bd4effadc2/udata_hydra/analysis/resource.py#L73), we only send the payload with `analysis:error` to udata and don't remove previous keys that may become irrelevant (`analysis:checksum`, `analysis:last-modified-at` etc.).

Indeed, extras update in udata has a dict `update` policy and a strip of keys with `None` value.

We should make sure to remove other analysis keys when facing an error by adding them in payload with None value.
Indeed, when a file has been analyzed but is now bigger than the limit, previous extras will be kept forever in udata.

On the contrary, on a successful analysis, we should make sure to remove eventual existing `analysis:error` by setting to None.

PR #256, opened 2025-03-26
Reset related udata metadata when sending

Pierlou:
Fix https://github.com/datagouv/hydra/issues/137

### DIFF
--- a/udata_hydra/utils/http.py
+++ b/udata_hydra/utils/http.py
@@ -16,8 +16,14 @@ class UdataPayload:
         "check": ["available", "date", "error", "status", "timeout"],
         "check:headers": ["content-type", "content-length"],
         "analysis": [
-            "checksum", "content-length", "error", "last-modified-at",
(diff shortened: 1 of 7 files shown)


udata_hydra/utils/http.py
```diff
@@ -11,6 +11,33 @@
...
+        "check:headers": ["content-type", "content-length"],
+        "analysis": [
+            "checksum",
+            "content-length",
+            "error",
+            "last-modified-at",
+            "last-modified-detection",
+            "mime-type",
+            "status",
+            "timeout",
```
maudetes, line 26: We don't have `status` and `timeout` in `analysis` namespace?

maudetes (approved):
Thank you! :pray:

udata_hydra/utils/http.py
Pierlou: You're totally right, I'm deleting them, thanks!

PR #257, opened 2025-03-31
Make sure detected_last_modified_at is set even if HAS_NOT_CHANGED

maudetes:
It is indeed used at each check to compare existing with previous value. If it isn't set, it will have to pursue the analysis instead of using the harvested metadata.
This leads to useless resource download if it can't use early hints (headers).

Example on [this resource](https://www.data.gouv.fr/api/2/datasets/resources/3f8ea372-3847-495d-9b93-c4a76d553166/), today we have detected_last_modified_at set only once out of twice
```
select created_at,detected_last_modified_at from checks where resource_id = '3f8ea372-3847-495d-9b93-c4a76d553166' ORDER BY created_at;
          created_at           | detected_last_modified_at  
-------------------------------+----------------------------
 2025-01-02 08:19:15.361935+01 | 2023-01-17 13:11:56.362+01
 2025-01-09 08:19:16.795927+01 | 
 2025-01-14 17:03:15.83802+01  | 2023-01-17 13:11:56.362+01
 2025-01-15 16:32:47.309478+01 | 
 2025-01-16 16:33:10.14949+01  | 2023-01-17 13:11:56.362+01
 2025-01-17 04:59:38.020002+01 | 
 2025-01-18 04:59:41.366839+01 | 2023-01-17 13:11:56.362+01
```

### DIFF
--- a/tests/test_crawl/test_crawl.py
+++ b/tests/test_crawl/test_crawl.py
@@ -359,6 +359,7 @@ async def test_change_analysis_harvested(
     res = await db.fetch("SELECT * FROM checks ORDER BY created_at DESC LIMIT 1")
     assert res[0]["detected_last_modified_at"].isoformat() == "2022-12-06T05:00:32.647000+00:00"
 
+
 @pytest.mark.catalog_harvested
 async def test_no_change_analysis_harvested(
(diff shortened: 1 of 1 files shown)


Pierlou (approved):
🚀

PR #258, opened 2025-04-02
Fill in new check even if no update

Pierlou:


udata_hydra/analysis/resource.py
```diff
@@ -108,6 +108,13 @@ async def analyse_resource(
                 },
             )
 
+    # if the file has not changed, update the current check with the infos of the previous one
+    if not dl_analysis and last_check:
```
maudetes, line 112: could we replace that by `elif last_check`?

maudetes (approved):

PR #259, opened 2025-04-04
Better change detections

Pierlou:


udata_hydra/analysis/resource.py
```diff
@@ -187,11 +187,21 @@ async def detect_resource_change_from_checksum(
...
+            return Change.HAS_CHANGED, {
+                "analysis:last-modified-at": datetime.now(timezone.utc).isoformat(),
+                "analysis:last-modified-detection": "computed-checksum",
+            }
+        else:
+            return Change.HAS_NOT_CHANGED, {
+                "analysis:last-modified-at": (
+                    last_check["last_modified"].isoformat()
+                    if last_check.get("last_modified")
+                    else None
```
Pierlou, line 201: I am not sure of this: we want the info to passed on from the previous check, but is it always there?

udata_hydra/analysis/resource.py
```diff
@@ -227,13 +237,22 @@ async def detect_resource_change_from_content_length_header(
...
+            "analysis:last-modified-at": changed_at.isoformat(),
+            "analysis:last-modified-detection": "content-length-header",
+        }
+    # same content_length is not 100% certainly no change, but a good tradeoff to prevent many downloads
+    return Change.HAS_NOT_CHANGED, {
+        # no change, using the last-modified from the previous check (passed on from check to check)
+        "analysis:last-modified-at": (
+            data[1]["detected_last_modified_at"].isoformat()
+            if data[1].get("detected_last_modified_at")
+            else None
```
Pierlou, line 252: Same here

maudetes (commented):
I think in a next iteration, we can store (and expose) the detection method since we're at it?

udata_hydra/analysis/resource.py
```diff
@@ -227,13 +237,22 @@ async def detect_resource_change_from_content_length_header(
     if len(data) <= 1 or not data[0]["content_length"]:
         return Change.NO_GUESS, None
     changed_at = data[0]["created_at"]
```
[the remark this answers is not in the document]
Pierlou: > If it isn't there, we'll act similarly as today?

Yes, so maybe the better thing to do is to return the payload if we have the info in the previous check, and otherwise return `(Change.NO_GUESS, None)` so that we catch the info downstream (with a potential download)?

> Maybe add a comment on changed_at? Shouldn't it be set in the HAS_CHANGED case only?

You're right, I'll move it inside the `if`

udata_hydra/analysis/resource.py
```diff
@@ -187,11 +187,21 @@ async def detect_resource_change_from_checksum(
...
+    if last_check:
+        if last_check.get("checksum") != new_checksum:
+            return Change.HAS_CHANGED, {
+                "analysis:last-modified-at": datetime.now(timezone.utc).isoformat(),
+                "analysis:last-modified-detection": "computed-checksum",
+            }
+        else:
+            return Change.HAS_NOT_CHANGED, {
+                "analysis:last-modified-at": (
+                    last_check["last_modified"].isoformat()
```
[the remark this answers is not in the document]
Pierlou: Yes you're totally right I mixed them up here, `last_modified` is from the headers of the previous check. I'll change it!

udata_hydra/analysis/resource.py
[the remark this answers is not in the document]
maudetes: Let's do that!

udata_hydra/analysis/resource.py
```diff
@@ -187,11 +187,18 @@ async def detect_resource_change_from_checksum(
         "analysis:last-modified-detection": "computed-checksum",
     }
     """
-    if last_check and last_check.get("checksum") != new_checksum:
-        return Change.HAS_CHANGED, {
-            "analysis:last-modified-at": datetime.now(timezone.utc).isoformat(),
-            "analysis:last-modified-detection": "computed-checksum",
-        }
+    if last_check:
```
maudetes, line 190: do we need to make sure last_check checksum is not None?
```suggestion
    if last_check and last_check.get("checksum"):
```

udata_hydra/analysis/resource.py
```diff
@@ -226,14 +233,21 @@ async def detect_resource_change_from_content_length_header(
     # content-length variation between current and last check
     if len(data) <= 1 or not data[0]["content_length"]:
         return Change.NO_GUESS, None
-    changed_at = data[0]["created_at"]
-    payload = {
-        "analysis:last-modified-at": changed_at.isoformat(),
-        "analysis:last-modified-detection": "content-length-header",
-    }
     if data[0]["content_length"] != data[1]["content_length"]:
```
maudetes, line 236: same, do we need to make sure content_length aren't `None`?

udata_hydra/analysis/resource.py
Pierlou: Next line is `if last_check.get("checksum") != new_checksum:`, is it included?

udata_hydra/analysis/resource.py
Pierlou: True! I guess testing only one shoud be enough?

udata_hydra/analysis/resource.py
maudetes: what if `last_check.get("checksum")` is `None` and `new_checksum` is `9985dbbf352fa2bb4b49b517ec95b4edfd1f1baf`, the comparison isn't fair?

Pierlou: If we have a `new_checksum` then it means we have downloaded the file, but last time we didn't, so I'd say very little chance they're the same? But it's cleaner anyway, I'll add that

PR #260, opened 2025-04-15
Geojson to pmtiles

hacherix:


udata_hydra/utils/file.py
```diff
@@ -77,3 +78,15 @@ async def download_resource(
         ]:
             tmp_file = read_csv_gz(tmp_file.name)
         return tmp_file
+
+
+def get_filename_from_path(file_path: str) -> str:
```
hacherix, line 83: You can delete this function it is a leftover from a first iteration.

hacherix (commented):
Well done! 🚀 🗺️ 🛰️

udata_hydra/analysis/resource.py
```diff
@@ -70,7 +72,8 @@ async def analyse_resource(
     change_status, change_payload = await detect_resource_change_on_early_hints(resource)
 
     # could it be a CSV? If we get hints, we will analyse the file further depending on change status
-    is_tabular, file_format = await detect_tabular_from_headers(check)
+    is_tabular, file_format = detect_tabular_from_headers(check)
+    is_geojson: bool = detect_geojson_from_headers(check)
     max_size_allowed = None if exception else int(config.MAX_FILESIZE_ALLOWED[file_format])
```
maudetes, line 77: `file_format` is not set if is geojson?

udata_hydra/analysis/resource.py
```diff
@@ -70,7 +72,8 @@ async def analyse_resource(
     change_status, change_payload = await detect_resource_change_on_early_hints(resource)
 
     # could it be a CSV? If we get hints, we will analyse the file further depending on change status
```
maudetes, line 74:
```suggestion
    # could it be a CSV or a GeoJSON? If we get hints, we will analyse the file further depending on change status
```

maudetes (commented):
So cool! :star_struck: 
It's actually quite separated from csv analysis, well done!

udata_hydra/utils/geojson.py
```diff
@@ -0,0 +1,11 @@
+import json
+
+
+async def detect_geojson_from_headers(check: dict) -> bool:
+    headers: dict = json.loads(check["headers"] or "{}")
+    if any(
+        headers.get("content-type", "").lower().startswith(ct)
+        for ct in ["application/vnd.geo+json"]
```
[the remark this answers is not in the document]
Pierlou: Hmm that means we will test quite more files, and fail in most cases 🤔 we need to have a safeguard before attempting to convert the file into pmtiles to prevent unelegant crashes

udata_hydra/analysis/resource.py
Pierlou: Indeed, thanks for spotting this 🙏

udata_hydra/utils/geojson.py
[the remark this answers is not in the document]
maudetes: Yes I think, we could add a check on the url (we've had the case with `csv.gz` before I think)

maudetes (approved):

Issue #247, opened 2025-02-26
Polish parquet exposure


Pierlou:
- [ ] when a resource is changed and reexposed as parquet, the previous parquet file is not deleted (I think?)
- [ ] as the name of the parquet file changes when the file is changed, it's not possible to use the URL to the parquet file on Minio as a stable URL (one has to fetch the `parquet_url` in the extras)

PR #262, opened 2025-04-22
Parquet files are named after resource ids

Pierlou:
This solves two issues (https://github.com/datagouv/hydra/issues/247):
- we don't need to come up with a cleaning mechanism to handle when a new parquet file is created (currently the previous parquet file is not deleted, so they'll just stack up endlessly)
- the parquet URL becomes stable `https://object.files.data.gouv.fr/hydra-parquet/hydra-parquet/{resource_id}.parquet`

We can then migrate the current parquet files (or wait until they're all re-generated)


tests/test_parquet_export.py
```diff
@@ -72,8 +75,10 @@ async def execute_csv_to_parquet() -> tuple[str, int] | None:
...
 
     else:
-        # TODO: don't use the exception as the assertion, better to mock the minio client sending the file
-        with pytest.raises(ValueError, match="invalid bucket name"):
-            await execute_csv_to_parquet()
+        url = "http://minio/test.parquet"
+        with patch("udata_hydra.analysis.csv.minio_client.send_file", return_value=url):
+            parquet_url, parquet_size = await execute_csv_to_parquet()
+        assert parquet_url == url
+        assert isinstance(parquet_size, int)
```
maudetes, line 82: maybe we can assert the resource_id is in parquet_url? :package:

maudetes (approved):
:pray:

tests/test_parquet_export.py
Pierlou: Not really as it stands because we are mocking the whole `send_file` function which is responsible for creating the URL. I have not been able to mock only the `fput_object` function (which is the one that should *really* be mocked, the rest should be tested too. My approach:
```python
        mocked_minio = MinIOClient()
        mocked_minio.client = MagicMock()
        mocked_minio.client.fput_object.return_value = None
        with patch("udata_hydra.analysis.csv.minio_client", return_value=mocked_minio):
            parquet_url, parquet_size = await execute_csv_to_parquet()
```
But while `mocked_minio` is of type `MinIOClient` and `mocked_minio.client` is a `MagicMock`, but within `csv_to_parquet`, `minio_client` becomes of type `MagicMock`... I'm taking any advice 🙏

maudetes (approved):
:pray:

PR #264, opened 2025-05-07
Update code for resources_exceptions table migration

Pierlou:
⚠️ this should be merged when the table is manually moved from the `main` db to the `csv` db (for `postgrest` access of indexes for `tabular-api`)


maudetes (approved):
we're good to go I think :rocket:

PR #266, opened 2025-05-13
Validate then check if resource has been analyzed before

Pierlou:
⚠️ do not merge before a true `csv-detective` is released and put into pyproject
If a tabular resource has been analyzed before, we check if the previously detected formats are still valid, which cuts much processing time compared to a full analysis (if successful of course). For instance for [this resource](https://www.data.gouv.fr/fr/datasets/6569b4473bedf2e7abad3b72/#/resources/00cf4398-e57f-419b-96ea-671cf9a8500e) from Météo France, which can be updated daily, we go from ~1000s down to ~300s for the whole process


udata_hydra/analysis/csv.py
```diff
@@ -183,6 +195,28 @@ async def analyse_csv(
         await Resource.update(resource_id, {"status": None})
 
 
+async def get_previous_analysis(resource_id: str) -> dict | None:
+    db = await context.pool("csv")
+    q = (
+        "SELECT parsing_table, csv_detective FROM tables_index "
+        f"WHERE resource_id='{resource_id}' ORDER BY created_at DESC"
```
maudetes, line 202:
```suggestion
        f"WHERE resource_id='{resource_id}' ORDER BY created_at DESC LIMIT 1"
```

udata_hydra/analysis/csv.py
```diff
@@ -183,6 +195,28 @@ async def analyse_csv(
...
+    )
+    res = await db.fetch(q)
+    if not res:
+        return None
+    analysis = json.loads(res[0]["csv_detective"])
+    # the csv_detective column is JSONB, so keys are reordered compared to the actual table
```
maudetes, line 208: I guess csvdetective shouldn't fail validation if keys are reordered? Or were they ordered dict?

udata_hydra/utils/errors.py
```diff
@@ -61,6 +61,7 @@ async def handle_parse_exception(
     if table_name is not None:
         db = await context.pool("csv")
         await db.execute(f'DROP TABLE IF EXISTS "{table_name}"')
+        await db.execute(f"DELETE FROM tables_index WHERE parsing_table='{table_name}'")
```
maudetes, line 64: :eyes:

maudetes (commented):
Nice, it's quite easy to understand the logic here :)

udata_hydra/analysis/csv.py
Pierlou: It's not for `csv-detective` (which indeed doesn't require the columns to be properly ordered) but for `smart_cast`:
```python
smart_cast(t, v, failsafe=True) for t, v in zip(columns.values(), line)
```

udata_hydra/utils/errors.py
Pierlou: We should do that right?

maudetes: ahaha exactly my question! I wonder if we should *mark them* as deleted? I wasn't sure of the implications

Pierlou: To me it makes no sense to keep the row in `tables_index` if the table doesn't exist anymore, it won't be used anyway will it?

maudetes (approved):
Looks nice! :clap:

PR #267, opened 2025-05-16
Try using tippecanoe python lib

maudetes:
Instead of using subprocess, we use tippecanoe python lib, that should use correct bin path for tippecanoe

See the content of `/lib/python3.11/site-packages/tippecanoe/__init__.py`


Pierlou (approved):
Clean, thanks for the fix 🤞

PR #273, opened 2025-06-03
Select DISTINCT parsing table in purge_selected_csv_tables CLI

maudetes:
Without this, we end up trying to delete the same parsing_table mutliple times.
It doesn't fail, but it wastes useless time deleting already deleted contents


maudetes:
`csv_to_db_index` inserts new rows but does not overwrite rows that may exists

maudetes:
> Is that what we want, or should we also prevent this behaviour? Why would we want to keep (indefinitely) the infos for tables that do not exist anymore or have been refilled?

I think it's interested to keep the previous db-ification information, however we could have a purge strategy as is done for checks :+1:

PR #279, opened 2025-06-19
Add parsing_table in extra analysis payload

maudetes:
It is useful for cdata to know specifically which are available between geojson, pmtiles & DB table.

We should probably migrate existing extras to add the `parsing_table` extras on the stock.


tests/test_analysis/test_analysis_csv.py
```diff
@@ -373,6 +374,7 @@ async def test_forced_analysis(
         webhook = rmock.requests[("PUT", URL(udata_url))][0].kwargs["json"]
         assert webhook.get("analysis:parsing:started_at")
         assert webhook.get("analysis:parsing:finished_at")
+        assert webhook.get("analysis:parsing:parsing_table")
         assert webhook.get("analysis:parsing:error") is None
         for k in ["parquet_size", "parquet_url"]:
             assert webhook.get("analysis:parsing:error", False) is None
```
Pierlou, line 380:
```suggestion
            assert webhook.get(f"analysis:parsing:{k}", False) is None
```

PR #287, opened 2025-07-10
Better Sentry logs

Pierlou:


udata_hydra/analysis/csv.py
```diff
@@ -139,7 +139,11 @@ async def analyse_csv(
                 )
         except Exception as e:
             raise ParseException(
-                step="csv_detective", resource_id=resource_id, url=url, check_id=check["id"]
+                message=e.__repr__(),
```
bolinocroustibat, line 142: I think it actually should be `str(e)`. `str(e)` returns the human-readable error message, while `e.__repr__()` returns a more technical, debug-oriented representation. Example with ValueError("bad input"):
	•	str(e) → "bad input"
	•	repr(e) → "ValueError('bad input')"

bolinocroustibat (changes requested):

udata_hydra/analysis/csv.py
Pierlou: I wouldn't mind having the error type as well, as you want!

bolinocroustibat (approved):
Titanic Ship

PR #290, opened 2025-07-15
ci: simplify ci

bolinocroustibat:
## Simplify CI configuration

Taking some inspiration from [this](https://github.com/datagouv/api-tabular/pull/50) and [this](https://github.com/datagouv/api-tabular/pull/38) recent CI refactoring PRs of TabularAPI, this PR streamlines the CircleCI configuration by:

- **Switching to Docker images which include Poetry 2** instead of manually installing Poetry in each job
- **Simplifying dependency installation** - using `poetry install` directly instead of creating a local Poetry venv
- **Using `-slim` variants of Docker images** for jobs that don't need full system dependencies (install, lint, build, publish)
- **Using full Python image for tests** where system dependencies like `libmagic` is required
- **Using flexible Python version** - `3.11` instead of `3.11.10` for automatic patch updates

### Benefits

- **Reduced complexity**: Fewer manual setup steps and custom Poetry installation
- **Faster execution**: Tests job appears to run faster with the new base image
- **Better maintainability**: Standardized Docker images with pre-installed tools
- **Smaller image sizes**: Using `-slim` variants where possible reduces container overhead

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2.1
 parameters:
   python-version:
     type: string
-    default: "3.11.2"
+    default: "3.11"
   publish-branch:
     type: string
     default: "main"
@@ -21,7 +21,7 @@ parameters:
 jobs:
   install:
     docker:
-      - image: cimg/python:<< pipeline.parameters.python-version >>
+      - image: acidrain/python-poetry:<< pipeline.parameters.python-version >>-slim
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -34,29 +34,22 @@ jobs:
...
+            poetry install --extras "dev"
       - save_cache:
           key: << pipeline.parameters.cache-prefix >>-{{ arch }}-{{ checksum "poetry.lock" }}
           paths:
             - .venv
-            - .poetry-venv
       - persist_to_workspace:
           root: .
           paths:
             - .
```
Pierlou, line 48: Do we still need this?

PR #298, opened 2025-07-23
Add more detailed statuses

Pierlou:


bolinocroustibat (approved):
Looks good to me, but:
- Where is VALIDATING_CSV being used?
- We need to be careful to monitor the preprod/prod queues for resources still with a deprecated statuses after the deploy. Would they still get blocked in a deprecated status?

bolinocroustibat:
> * `VALIDATING_CSV` is used [here](https://github.com/datagouv/hydra/pull/298/files#diff-c1ba41ef586f9cb81ea0c11340a336c5c46ee4686cb4dc27698a06a78483d34f)
> * we now have a [status resetting mechanism](https://github.com/datagouv/hydra/blob/f7928846f0320635def2667d05a6a2d56b51c42d/udata_hydra/cli.py#L124) that prevents stucked resources everyday

Amazing!

PR #301, opened 2025-07-25
Fix crawler endpoint

Pierlou:


bolinocroustibat (approved):
Thanks for cleaning this and spotting this rare error!

PR #303, opened 2025-07-25
Handle new geo column formats from csv-detective for geojson

Pierlou:
Following https://github.com/datagouv/csv-detective/pull/133, we now handle:
- `[lat,lon]` (with brackets and potential blank spaces)
- `lon,lat` and `[lon,lat]` a completely new format (mainly to match [TableSchema's geopoint](https://specs.frictionlessdata.io//table-schema/#geopoint))


bolinocroustibat (approved):
LGTM.

Out of curiosity, are the lon,lat and [lon,lat] formats common? Where is it used?
I always thought latitude was always before longitude, out of convention.

bolinocroustibat:
> I haven't been able to find whether there is a true consensus on the order of the coordinates, apparently [both ways coexist](https://fr.wikipedia.org/wiki/GeoJSON#:~:text=en%20fonction%20de%20certaines%20impl%C3%A9mentations) 😕 for instance in TableSchema's [geopoint object](https://specs.frictionlessdata.io//table-schema/#geopoint), longitude comes first, so all IRVE resources have this syntax (see [this one](https://www.data.gouv.fr/datasets/fichier-irve-gireve/#/resources/61387a4e-22f7-4662-b241-d5cac4dd91fd))

Interesting!

PR #304, opened 2025-07-25
docs: fix README typos

bolinocroustibat:
- Fix typos and example commands in README
- While we're at it, add emojis to sections of README, because... why not


Pierlou (approved):
Thanks for the clean up!

PR #305, opened 2025-07-25
chore: upgrade dependencies and fix pytest-asyncio deprecated code

bolinocroustibat:
Upgrade dependencies to their latest versions, including update of `pytext-asyncio` from 0.25.3 to 1.1.0.
`pytest-asyncio` >= 1.0 has some breaking changes, including the deprecation of `event_loop`, which makes the testing of async features more easy and less verbose, but need some code adaptation to remove `event_loop`.

Detail:
https://pytest-asyncio.readthedocs.io/en/latest/reference/changelog.html
https://github.com/pytest-dev/pytest-asyncio/pull/1106


Pierlou (approved):
Thanks for the upgrades 🚀

PR #307, opened 2025-08-12
Add check id to udata extras to add insights

Pierlou:
As the sending of the `check:...` extras is done quite before the one related to `analysis:...` (and in a separate thread), it can happen (and as of now it indeed happens quite often) that the two sets of extras are from a different check, inducing confusion (hydra has done `check_resource`, but not `analyse_resource`, which is still in queue). These additions allow to track how late hydra is

### DIFF
--- a/udata_hydra/utils/http.py
+++ b/udata_hydra/utils/http.py
@@ -13,12 +13,13 @@ log = logging.getLogger("udata-hydra")
 
 class UdataPayload:
     HYDRA_UDATA_METADATA = {
-        "check": ["available", "date", "error", "status", "timeout"],
+        "check": ["available", "date", "error", "id", "status", "timeout"],
         "check:headers": ["content-type", "content-length"],
         "analysis": [
             "checksum",
             "content-length",
             "error",
+            "check_id",
             "last-modified-at",
             "last-modified-detection",
             "mime-type",
--- a/udata_hydra/analysis/resource.py
+++ b/udata_hydra/analysis/resource.py
@@ -130,7 +130,7 @@ async def analyse_resource(
             has_changed=change_status == Change.HAS_CHANGED,
         )
 
-    analysis_results = {**dl_analysis, **(change_payload or {})}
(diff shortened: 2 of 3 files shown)


udata_hydra/db/check.py
```diff
@@ -107,17 +107,17 @@ async def insert(cls, data: dict, returning: str = "id") -> dict:
 
         Note: Returns dict instead of Record because this method performs additional operations beyond simple insertion (joins with catalog table, adds dataset_id).
         """
-        data = convert_dict_values_to_json(data)
-        q1: str = compute_insert_query(table_name="checks", data=data, returning=returning)
+        json_data = convert_dict_values_to_json(data)
```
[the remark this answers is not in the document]
Pierlou: It was replacing the source data which was causing trouble downstream (and wasn't very clean imo)

bolinocroustibat:
> > Is there a test to test the data sent to udata?
> 
> Currently we don't, [all the calls to udata are mocked](https://github.com/datagouv/hydra/blob/2ed675fd6542165f09e4b8c89523e1a86ce37a8e/tests/conftest.py#L166), maybe it's a good opportunity to make some

That'd be great but since there were none anyway before the PR, it shouldn't be a blocker to merge this PR

udata_hydra/utils/http.py
```diff
@@ -13,12 +13,13 @@
 
 class UdataPayload:
     HYDRA_UDATA_METADATA = {
-        "check": ["available", "date", "error", "status", "timeout"],
+        "check": ["available", "date", "error", "id", "status", "timeout"],
         "check:headers": ["content-type", "content-length"],
         "analysis": [
```
maudetes, line 18: Do we want to add an analysis date?

Pierlou: We have `analysis:parsing:finished/finished_at`, do we want something on top?

maudetes: Hmm true! I think it's the subsequent task for csv/geojson only and not the generic resource analysis, but we probably don't need it.

Issue #306, opened 2025-08-11
Issues with Sentry


Pierlou:
We still have some issues with the communication between hydra, Sentry and udata:
- we often end up with no StackTrace in Sentry, making it hard to trace the errors to their roots (see [this](https://errors.data.gouv.fr/organizations/sentry/issues/261186/?environment=hydra&project=21&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=2) and [this](https://errors.data.gouv.fr/organizations/sentry/issues/261195/?environment=hydra&project=21&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=7d&stream_index=4) for instance, we can investigate using the queries, but not with a clear exception chain in the code)
- we often end up with no Sentry id in the `parsing_error` metadata in udata (see [this case](https://crawler.data.gouv.fr/api/checks/all/?resource_id=51373445-efb4-485c-84f3-5b31ec6de67d), the error is [not pushed to udata](https://www.data.gouv.fr/api/2/datasets/resources/51373445-efb4-485c-84f3-5b31ec6de67d/) btw)

These two cases might be related, or not 🕵️

PR #308, opened 2025-08-14
fix: fix custom Sentry error to include stack trace

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/306


udata_hydra/utils/errors.py
```diff
@@ -36,9 +36,21 @@ def __init__(
...
+                # Automatically retrieve the original exception from exception chaining
+                # __cause__ is set with 'raise ... from e', __context__ is set automatically
+                original_exception = getattr(self, "__cause__", None) or getattr(
+                    self, "__context__", None
+                )
+                if original_exception:
+                    # Capture the ORIGINAL exception with its full stack trace
+                    sentry_sdk.capture_exception(original_exception)
+                else:
+                    # Fallback: capture self if no original exception
```
Pierlou, line 52: I'd say this can never happen as we always raise `these types of exceptions when the inner processes fail, from the exception that occurs, but I guess we never know 🤞

tests/test_utils/test_sentry_error.py
Pierlou, line 1: Thanks for adding a dedicated test 🙏

Pierlou (approved):
Big big thanks for digging into this 🫶

udata_hydra/utils/errors.py
bolinocroustibat: Yes. If we forget to do `...from e`, maybe?

Pierlou: There used to be cases where we were missing this syntax, but I think we fixed them all. Maybe we could have a test for this? But that could be hard to implement

Pierlou: Sounds fair, if we end up seeing one such case, we'll know where to look 🔍

PR #312, opened 2025-08-19
feat: allow crawl of non datagouv URLs with CLI

bolinocroustibat:


maudetes (commented):
I wonder if we shouldn't have a cli that takes a URL and add a new entry in catalog (similarly to `insert-resource-into-catalog`). And if users run `analyse_csv` with an unknown URL, we can tell them that they should first import it, giving them the command.

I think it would prevent modifying the `csv.py` that is already quite complex.

PR #313, opened 2025-08-20
feat: add resource_id info in timer logs and add timer for analysing …

bolinocroustibat:
…resource


Pierlou (approved):
Thanks, we'll appreciate next time we dig into the logs 🕵️

PR #315, opened 2025-08-21
fix: fix ExceptionWithSentryDetails to include Python stack trace

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/306

Fix custom exception which enriches Sentry with tags and captures the exception with full stack trace when it's actually raised and caught.

## Why the original implementation failed

**Wrong timing - auto-capture in `__init__`**  
- The original implementation captured the exception in the `__init__` method  
- At this point, the exception hasn't been raised yet, so there's no stack trace  
- Sentry receives an exception without any context about where it occurred  

## Why this solution works

 We now capture the exception in the `__str__` method . This method is called when the exception is converted to string  which happens after the exception is raised and has a full stack trace . Sentry receives the complete context: exception + stack trace + custom tags  

## Another attempt which didn't work as we wanted it to

We also tried another implementation with manual enrichment where tags were set in a separate scope:  

```python
with sentry_sdk.new_scope() as scope:
    scope.set_tags({...})  # Tags are set
    # Scope ends here and tags are lost!

# Later...
sentry_sdk.capture_exception(e)  # No tags available!
```

But the scope was temporary and didn't remain active during the entire capture process , so tags were lost.

### DIFF
--- a/udata_hydra/utils/errors.py
+++ b/udata_hydra/utils/errors.py
@@ -68,7 +68,7 @@ class ExceptionWithSentryDetails(Exception):
         This happens when it's raised and caught, ensuring full stack trace.
         """
         # Only capture if we have a traceback (exception is being raised)
-        if hasattr(self, "__traceback__") and self.__traceback__ is not None:
+        if getattr(self, "__traceback__", None) is not None:
             if sentry_sdk.Hub.current.client:
                 with sentry_sdk.new_scope() as scope:
                     scope.set_tags(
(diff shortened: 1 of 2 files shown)


udata_hydra/utils/errors.py
```diff
@@ -27,32 +55,46 @@ def __init__(
...
+        # NO auto-capture here - let the exception be raised normally
         super().__init__(message, *args)
 
+    def __str__(self):
+        """
+        Capture to Sentry when the exception is converted to string.
+        This happens when it's raised and caught, ensuring full stack trace.
+        """
+        # Only capture if we have a traceback (exception is being raised)
+        if hasattr(self, "__traceback__") and self.__traceback__ is not None:
```
Pierlou, line 71: NIT
```suggestion
        if getattr(self, "__traceback__", None) is not None:
```

Pierlou (approved):
Thanks for digging into this again and finding the plot hole 💪

PR #317, opened 2025-08-22
refactor: split `csv_to_geojson_and_pmtiles()` into intermediary function `csv_to_geojon()` for tests

bolinocroustibat:
Partly helps https://github.com/datagouv/data.gouv.fr/issues/1847:

- Split `csv_to geojson_and_pmtiles()` function into a new intermediary function `csv_to_geojon()`, in order to better unit test and benchmark the csv to PMtiles workflow later ; so that functions `csv_to_geojon()` and `geojson_to_pmtiles()` can now be called by the tests or from the CLI independently with new, consistent arguments and return types for testing.

Also, on the way:

- Use of `pathlib.Path` instead of strings for file paths whenever possible in `csv_to geojson_and_pmtiles()`, `csv_to_geojon()` and `geojson_to_pmtiles()`, for safer, cleaner, and more readable file operations.

- Adapt tests to cleanup the test data files through a `cleanup` function argument instead of a mock on `os`.

The branch `benchmark` will be rebased on this PR, so measure performance of each step independently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@
 - Optimize queue priorities for resource processing [#311](https://github.com/datagouv/hydra/pull/311)
 - Add `resource_id` info in timer logs and add timer for analysing resource [#313](https://github.com/datagouv/hydra/pull/313)
 - Add check id in udata extras to facilitate debug [#307](https://github.com/datagouv/hydra/pull/307)
-- Split `csv_to geojson_and_pmtiles()` function into a new intermediary function `csv_to_geojson()` for better unit testing/benchmarking [#317](https://github.com/datagouv/hydra/pull/317)
+- Split `csv_to geojson_and_pmtiles` function into a new intermediary function `csv_to_geojson` for better unit testing/benchmarking [#317](https://github.com/datagouv/hydra/pull/317)
 
 ## 2.3.0 (2025-07-15)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Optimize queue priorities for resource processing [#311](https://github.com/datagouv/hydra/pull/311)
 - Add `resource_id` info in timer logs and add timer for analysing resource [#313](https://github.com/datagouv/hydra/pull/313)
 - Add check id in udata extras to facilitate debug [#307](https://github.com/datagouv/hydra/pull/307)
+- Split `csv_to geojson_and_pmtiles()` function into a new intermediary function `csv_to_geojson()` for better unit testing/benchmarking [#317](https://github.com/datagouv/hydra/pull/317)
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Optimize queue priorities for resource processing [#311](https://github.com/datagouv/hydra/pull/311)
 - Add `resource_id` info in timer logs and add timer for analysing resource [#313](https://github.com/datagouv/hydra/pull/313)
 - Add check id in udata extras to facilitate debug [#307](https://github.com/datagouv/hydra/pull/307)
+- Split `csv_to geojson_and_pmtiles()` function into a new intermediary function `csv_to_geojson()` for better unit testing/benchmarking [#317](https://github.com/datagouv/hydra/pull/317)
```
Pierlou, line 20: NIT 🤓 
```suggestion
- Split `csv_to geojson_and_pmtiles` function into a new intermediary function `csv_to_geojson` for better unit testing/benchmarking [#317](https://github.com/datagouv/hydra/pull/317)
```

udata_hydra/analysis/geojson.py
```diff
@@ -100,55 +101,31 @@ async def analyse_geojson(
         await Resource.update(resource_id, {"status": None})
 
 
-async def geojson_to_pmtiles(
-    file_path: str,
+async def csv_to_geojson(
```
Pierlou, line 104: I'm realizing the function's name is a bit overlooked, at this point we're converting a DataFrame into geojson, not really a csv anymore (or even other file types), but as we're aiming to coming back to a real file conversion I guess it's fine

udata_hydra/analysis/geojson.py
```diff
@@ -207,7 +178,7 @@ def prevent_nan(value):
                 {
                     "type": "Feature",
                     # json is not pre-cast by csv-detective
-                    "geometry": json.loads(row[geo["geometry"]]),
+                    "geometry": json.loads(str(row[geo["geometry"]])),
```
Pierlou, line 181: I am not sure why we need to convert the field to string, don't we explicitly pass `cast_json=False` to `csv-detective`?

bolinocroustibat: AFAIK, Pandas has its own type system that extends Python's built-in types. The The `row[geo["geometry"]` value might be a pandas string type (like pandas.StringDtype), not a Python `str` type.
I guess doing a `str(pandas.StringDtype)` is useless, it was more to satisfy IDEs type checkers. Do you think it could have a micro-impact on performance? Can remove

Pierlou: `csv-detective` loads all columns with `dtype="unicode"` (I am considering moving to just `dtype=str` btw), which yields pure python strings in the cells, so I'd say it's 100% superfluous 🤔

PR #322, opened 2025-08-26
Upgrade csv-detective

Pierlou:
the `default=str` for `json.dumps` is not necessary, but to be on the safe side


maudetes (approved):
:+1:

bolinocroustibat (approved):
LGTM!

PR #325, opened 2025-08-29
perf: reuse HTTP connexions between sessions

bolinocroustibat:
## Reuse HTTP connexions between sessions

Refactors the HTTP client usage so that the application relies on a shared aiohttp.ClientSession instead of creating a new session for every request. This should reduce connection-setup overhead and enable connection pooling. Even though we currently don’t make many consecutive requests to the same host — so the impact will be limited for now — using a shared client keeps things consistent, more readable, and prepares us for scenarios where reuse matters.

**Changes:**
	•	Introduce centralized HTTP client management in `utils/http.py`.
	•	Replace direct ClientSession instantiations with calls to `get_http_client()`.

**Benchmarks:**
See [benchmarks in this PR on another repo](https://github.com/datagouv/csv-detective/pull/156). Additional benchmarking with many successive requests to the same server would better quantify the performance gains.


udata_hydra/utils/file.py
```diff
@@ -71,18 +72,15 @@ async def download_resource(
     i = 0
     too_large, download_error = False, None
     try:
-        async with aiohttp.ClientSession(
-            headers={"user-agent": config.USER_AGENT_FULL},
-            raise_for_status=True,
```
ThibaudDauce, line 76: The new version lose `raise_for_status=True`?

udata_hydra/utils/http.py
```diff
@@ -96,16 +98,58 @@ async def send(dataset_id: str, resource_id: str, document: UdataPayload) -> Non
...
+        _http_client = aiohttp.ClientSession(
+            timeout=timeout_obj,
+            headers=headers,
+        )
+        log.debug("Created new aiohttp ClientSession")
+
+    return _http_client
+
+
+async def close_http_client():
```
ThibaudDauce, line 148: Never used?

udata_hydra/utils/http.py
```diff
@@ -96,16 +98,58 @@ async def send(dataset_id: str, resource_id: str, document: UdataPayload) -> Non
...
+    """
+    global _http_client
+
+    if _http_client is None or _http_client.closed:
+        # Create a new client session
+        timeout_obj = aiohttp.ClientTimeout(total=timeout) if timeout else None
+
+        # Prepare headers
+        headers = {}
+        if config.USER_AGENT_FULL:
```
ThibaudDauce, line 136: Is this always true?

udata_hydra/utils/http.py
```diff
@@ -96,16 +98,58 @@ async def send(dataset_id: str, resource_id: str, document: UdataPayload) -> Non
...
+            raise IOException(
+                "Resource has been deleted on udata", resource_id=resource_id, url=uri
+            )
+        if resp.status == 502:
+            raise IOException("Udata is unreachable", resource_id=resource_id, url=uri)
+        else:
+            resp.raise_for_status()
+
+
+async def get_http_client(
```
ThibaudDauce, line 116: Is this function really `async`?

udata_hydra/utils/http.py
ThibaudDauce, line 96: Do we need this?

udata_hydra/utils/http.py
```diff
@@ -96,16 +98,58 @@ async def send(dataset_id: str, resource_id: str, document: UdataPayload) -> Non
...
+    Args:
+        follow_redirects: Whether to follow redirects
+        timeout: Request timeout in seconds
+
+    Returns:
+        Shared aiohttp ClientSession instance
+    """
+    global _http_client
+
+    if _http_client is None or _http_client.closed:
```
ThibaudDauce, line 130: Nothing ever close the connection?

ThibaudDauce (changes requested):
Have you thought about using context instead of global variable? I think it may be better to replicate what is done for the database pool in `context` /  `start_checks()`, no?

udata_hydra/utils/http.py
bolinocroustibat: Yes, this was added [here](https://github.com/datagouv/hydra/pull/328) for consistency and debugging

ThibaudDauce: But the new session system is adding it below for every requests, no?

bolinocroustibat:
Closing this, it won't help much in practice.

The crawler already reuses a ClientSession for each batch, then sleeps 60s, so keep-alive connections are gone before the next batch. RQ workers fork per job, so a shared client does not survive across send / download jobs either. Most requests also hit different hosts, so connection pooling would barely apply.

The change also has real issues (lost raise_for_status, unused close, global instead of context, ignored timeout/redirect args). Not worth merging as a performance fix. We can revisit later with a proper lifecycle in context if we actually need a long-lived client (e.g. udata callbacks).

PR #328, opened 2025-09-01
feat: dynamic version in user agent

bolinocroustibat:
- Fix the wrong version that was sent in user agent headers, using the app version
- Add user-agent key in requests sent to udata

That might help debugging some requests when received by external services later on.


udata_hydra/__init__.py
```diff
@@ -49,5 +49,12 @@ def __getattr__(self, __name):
     def __dict__(self):
         return self.configuration
 
+    @property
+    def USER_AGENT_FULL(self) -> str | None:
+        """Build the complete user agent string with version"""
+        if self.USER_AGENT and self.APP_VERSION:
+            return f"{self.USER_AGENT}/{self.APP_VERSION}"
+        return None
```
maudetes, line 57: Could we always return a value instead?

maudetes (approved):

PR #331, opened 2025-09-02
fix: fix CRUD types returns and reduce DB queries when cleanup statuses

bolinocroustibat:
1. We have some consistency issues with types and (non-blocking) type errors in the code. To make the code more consistent and avoid type warnings, we suggest the following rules:

	•	In the business logic, checks should be handled as `dict` rather than `Record` whenever possible, since working with native Python objects greatly simplifies the manipulation logic.

	•	Since `asyncpg` returns `Record` objects when querying the DB, CRUD methods should return `Record` types whenever possible so they remain “pure” CRUD.

This PR updates the code to follow these rules, and introduces a `dict` option to `Check` CRUD methods avoid repeating the pattern `record = await Check.get()` followed by `check = dict(record)` multiple times, in line with the DRY principle.

We also add  `# type: ignore` when calling `Check` class methods using `as_dict`, as the IDE may not infer the type correctly for methods that can return multiple types.

2. Reduce DB queries when cleaning up statuses in `Resource.clean_up_statuses()` by using a single DB request instead of a loop.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -284,7 +284,7 @@ async def fake_check():
             "geojson_url": "https://example.org/file.geojson" if pmtiles_url else None,
             "geojson_size": 1024 if geojson_url else None,
         }
-        check: dict = await Check.insert(data=data, returning="*")
+        check: dict = await Check.insert(data=data, returning="*", as_dict=True)
         data["id"] = check["id"]
         if check.get("dataset_id"):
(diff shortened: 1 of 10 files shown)


maudetes (commented):
I did not look a the code in details, but it seems to me that we're adding some complexity here when the for a less structured type?

PR #337, opened 2025-09-09
fix: fix user agent string

bolinocroustibat:
Fix the dynamic building of USER_AGENT_FULL since on the production setting we use USER_AGENT setting like "`udata-hydra/1.0 data.gouv.fr crawler <abuse@data.gouv.fr>`", and not "`udata-hydra`" or "`udata-hydra/1.0`" like in the default config.

We now use a regex so that it would work in all those cases:
- `udata-hydra/1.0 data.gouv.fr crawler <abuse@data.gouv.fr>` -> `udata-hydra/<APP_VERSION> data.gouv.fr crawler <abuse@data.gouv.fr>`
- `udata-hydra/X.X data.gouv.fr crawler <abuse@data.gouv.fr>` -> `udata-hydra/<APP_VERSION> data.gouv.fr crawler <abuse@data.gouv.fr>`
- `udata-hydra/1.0` -> `udata-hydra/<APP_VERSION>`
- `udata-hydra` -> `udata-hydra/<APP_VERSION>`
etc.

### DIFF
--- a/udata_hydra/__init__.py
+++ b/udata_hydra/__init__.py
@@ -1,6 +1,7 @@
 import importlib.metadata
 import logging
 import os
+import re
 import tomllib
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


maudetes (approved):
Thanks for the fix

PR #344, opened 2025-10-06
feat: separate database connection pools for each app function

bolinocroustibat:
Related to [this problem and discussion](https://mattermost.incubateur.net/betagouv/pl/hqsq5topgjygdxgp41fhwbpscy): attempt to solve the problem of limits reached for the number of postgresql connexion pools on production, by separating the connexion pool into 3 pools:
- worker (default)
- crawler
- API

So that an increase/max of connexions in one pool doesn't put another part of the app into jeopardy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -136,3 +136,6 @@ dmypy.json
 # Project specific
 config.toml
 tests/data/.*
+
+# vscode/codium
+.vscode/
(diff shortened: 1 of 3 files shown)


Pierlou (approved):
I haven't been able to check the conversation but the changes looks fine to me. Don't forget the changelog 🤓

PR #350, opened 2025-10-20
fix: fix entrypoint

bolinocroustibat:
- Fix entrypoint to avoid the udata-hydra not to be found (https://gitlab.com/etalab/data.gouv.fr/infra/-/jobs/11780906717). Explanation:

    - Poetry's approach: When Poetry sees udata_hydra.cli:run in [tool.poetry.scripts], if it finds run (even if it's an imported function) in udata_hydra.cli:run, it uses it. Poetry is more flexible about what constitutes a "callable" for entry points.

     - Standard setuptools/uv approach: When setuptools sees udata_hydra.cli:run in [project.scripts], it requires that run be a function defined in the udata_hydra.cli module's namespace.

- Update lock file

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -263,15 +263,6 @@ wheels = [
 ]
 
 [[package]]
-name = "bumpx"
-version = "0.3.10"
-source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


maudetes (approved):
:pray: thank you for the fix

PR #353, opened 2025-11-03
feat: add resource_id index on tables_index

maudetes:


Pierlou (approved):
I don't understand how we haven't done this sooner 🤯

PR #362, opened 2025-11-21
feat: add total resources and deleted resources in crawler status health check

bolinocroustibat:
- Add total resources and deleted resources in crawler status health check (to be later used in Munin dashboards)
- Reorganize crawler health check output to be more explicit


Pierlou (approved):
Nice to have more insights in the API!

PR #364, opened 2025-11-26
Add binary type support

Pierlou:
Fixes https://errors.data.gouv.fr/organizations/sentry/issues/270937/
To be added in `csv-detective` too


bolinocroustibat (approved):
Image

PR #371, opened 2026-01-07
chore: clean remaining pmtiles files

maudetes:
We were not cleaning pmtile files properly after upload to minio.

Additionally, tippecanoe was returning an exit code 104 if the output file was already existing, making the conversion fail.
We've added the [`--force` parameter](https://github.com/mapbox/tippecanoe?tab=readme-ov-file#output-tileset) to make sure it doesn't fail even if the ouptut file already exists.

## Next steps:
* remove existing pmtiles leftover on dev, demo & prod servers
* deploy hydra
* re-run geojson to pmtiles for resources with an error 104 on conversion


Pierlou (approved):
Thanks 🙏 🧹

PR #378, opened 2026-01-19
perf: use Python 3.14

bolinocroustibat:
Upgrading from Python 3.11 to 3.14 for performance gains on our pure-Python workloads.

## Benchmark results (circleci, same runner, 4 runs for Python 3.14)

| Test | 3.11 mean | 3.14 mean | Speedup |
|------|-----------|-----------|---------|
| CSV analysis | 50.5s | 29.2s | **1.73× (+42%)** |
| CSV → GeoJSON | 299.9s | 227.2s | **1.32× (+24%)** |
| PMTiles (tippecanoe, C++) | 212.5s | 207.8s | ~1× (expected) |

Results from the [benchmarks branch](https://github.com/datagouv/hydra/blob/benchmarks/.benchmarks/benchmarks.csv). The PMTiles test acts as a negative control (C++ tool, unaffected by Python version), confirming the CSV gains might be real.

The 4 Python 3.14 runs are consistent: 28.41s, 29.23s, 28.49s, 30.51s (standard deviation of 0.97s).

A single 3.13 run on a different runner showed no improvement over 3.11 — not conclusive given the different setup, but it hints the gains may be specific to 3.14 which is consistent with what we can read on [some performance reports](https://www.phoronix.com/review/python-314-benchmarks/2).

The speedup could come from Python 3.14's interpreter improvements, which benefit `csv-detective` and `dateparser` (pure-Python workloads)?


Pierlou (approved):
Well that was an easy review, thanks for the upgrade! We can try this on dev/demo ⚡

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "tippecanoe>=2.72.0",
     "typer>=0.15.1",
 ]
-requires-python = ">=3.11,<3.14"
+requires-python = ">=3.11,<3.15"
```
Pierlou, line 32: Should we upgrade the oldest supported version too? (depending on how the change in `conftest.py` is backwards-compatible)

bolinocroustibat: It is compatible with 3.11!

PR #391, opened 2026-02-18
clean: add integrated type checking

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/392.

Add [ty](https://docs.astral.sh/ty/) type checker to hydra to align with practices used in other backend stacks (e.g. [udata](https://github.com/opendatateam/udata/pull/3670)). This PR introduces it and resolves all reported type errors so that `ty check` passes.

- Add `ty` to dev dependencies and document `uv run ty check` in the README.
- Add a "Type check" step in the CircleCI lint job so type errors block the pipeline on tagged builds.
- Fix 100+ type checks issues

---

**EDIT:** Following review, code that was changed only for the type checker was reverted so behaviour matches `main` branch. Type checking still passes via targeted `# type: ignore`.

### DIFF
--- a/udata_hydra/utils/file.py
+++ b/udata_hydra/utils/file.py
@@ -29,7 +29,9 @@ def compute_checksum_from_file(filename: str) -> str:
 
 def extract_gzip(file_path: str) -> IO[bytes]:
     with gzip.open(file_path, "rb") as gz_file:
-        with tempfile.NamedTemporaryFile(mode="wb", delete=False) as temp_file:
+        with tempfile.NamedTemporaryFile(
+            dir=config.TEMPORARY_DOWNLOAD_FOLDER or None, mode="wb", delete=False
+        ) as temp_file:
             temp_file.write(gz_file.read())
     return temp_file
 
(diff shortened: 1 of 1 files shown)


bolinocroustibat:
> Thanks for introducing this! I've chekced most of the impacted files, and while it all seems clear to me, I have two inquiries:
> 
> * the code is _much_ heavier and harder to read, and sometimes even superfluously, for instance `csv_detective.routine` only returns a tuple if `output_df` is set to `True`, which is not the case here, so we can safely do `analysis = routine(...)` and use `analysis` as a dict
> * I feel like we ironically lose in safety as we introduce many `if ... is not None`-like checks: some functions can return an object or `None` and we use their output as not `None` because it's how things are expected. I'd rather have an error raised and fix an issue than to have the code silently skip cases where the output is unexpectedly `None`

I agree, this shouldn't make the code more difficult to read and maintain.
Let me see if we can add `type: ignore` to shut up ty where it is too picky and/or too dumb for those obvious cases, so we could have a type checking while not introducing complexity.

Pierlou (approved):
Thanks for the rework, this looks very clean to me 🚀

PR #401, opened 2026-03-30
feat(ogc): add support for WMS analysis

maudetes:
Building on https://github.com/datagouv/hydra/pull/385

The logic is largely the same between WFS and WMS.
The main difference are
- the outputFormat (that is only relevant for WFS)
- the crs_options (that are either Crs objects or raw strings)
- the main versions


Pierlou (approved):
I'm really not an expert of WFS/WMS but the logic looks good, thanks for making the existing functions modular, it helps a lot understand the flow 👏

udata_hydra/analysis/ogc.py
```diff
@@ -90,29 +111,36 @@ async def analyse_ogc(check: dict) -> OgcMetadata | None:
...
 
-                # Extract default CRS
+                # Extract default CRS (ie. the first CRS in the list)
                 crs_options = getattr(layer, "crsOptions", []) or []
                 if crs_options:
-                    ogc_layer["default_crs"] = crs_options[0].getcode()
+                    ogc_layer["default_crs"] = (
+                        crs_options[0].getcode()
+                        if isinstance(crs_options[0], Crs)
+                        else crs_options[0]
```
abulte: Is there a way _not_ to get a `Crs`?

abulte (commented):
💪 🙏

bolinocroustibat (approved):
I am very illiterate on the question of geo formats so my review is quite limited, but the rest looks good to me! Thank you

abulte (approved):

PR #402, opened 2026-04-01
 feat: generate parquet from PostgreSQL table instead of re-reading CSV 

ThibaudDauce:
On `MN_07_latest-2025-2026.csv`: **14s → 9s** (-35%)
On `ValeursFoncieres-2024.txt`: **69s → 17s** (-75%)
On `irve.csv`: **14s → 2s** (-85%)
On `joconde.csv`: **35s → 14s** (-60%)

### DIFF
--- a/tests/test_analysis/test_parquet_export.py
+++ b/tests/test_analysis/test_parquet_export.py
@@ -8,11 +8,12 @@ from csv_detective import routine as csv_detective_routine
 from tests.conftest import RESOURCE_ID
 from udata_hydra.analysis.csv import (
     RESERVED_COLS,
+    csv_to_db,
     csv_to_parquet,
     generate_records,
 )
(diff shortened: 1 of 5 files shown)


Pierlou:
> I think a side effect is indeed the casting that may differ (ie. converting "yes" to a boolean "true" for example (?)). I think we just need to clarify this bit, but in my opinion, it is justifiable to use the same casting approach between the SQL one and the other converted files (parquet, geojson & co).

Currently all converted files have the same casting logic (`generate_records` is called in all cases), so using the table as the source for other formats would not affect the files' contents

Pierlou (approved):
Thanks for the refactor in such a short amount of time!

bolinocroustibat (approved):
Well spotted! That might improve the perfs greatly. I might use the parquet in DB to create geoJSON and PMtiles later - but it should be always the case by default.

PR #403, opened 2026-04-02
feat: add detailed timer marks for geojson and pmtiles conversion steps

ThibaudDauce:
Split the opaque csv-to-geojson-pmtiles timer into two separate marks (csv-to-geojson and geojson-to-pmtiles) to identify bottlenecks.


Pierlou (approved):
Thanks, we'll be able to more accurately pinpoint the timings ⏲️

PR #404, opened 2026-04-02
feat: generate GeoJSON from PostgreSQL instead of re-reading CSV

ThibaudDauce:
Stream GeoJSON features directly from the database using a cursor, avoiding a third read of the CSV file. PostgreSQL builds the JSON with json_build_object, so no Python-level casting is needed.

Also extract _detect_geo_columns to deduplicate the geo column detection logic.

Will allow to send this part in another job.

On `MN_07_latest-2025-2026.csv`: **50s → 18s** (-64%)
On `joconde.csv`: **75s → 43s** (-42%)

Same as https://github.com/datagouv/hydra/pull/402 but for GeoJSON


udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+            'geometry', {geometry_sql},
+            'properties', {properties_sql}
+        )::text
+        FROM {_quote_ident(table_name)}
+        {where}
+    """
+    return query, params
+
+
+async def csv_to_geojson_from_db(
```
Pierlou: We may want to keep the function name close to its parquet counterpart
```suggestion
async def save_as_geojson_from_db(
```

udata_hydra/analysis/geojson.py
```diff
@@ -154,33 +170,22 @@ def get_features(
                     "properties": {col: row[col] for col in row.keys() if col != geo["geojson"]},  # type: ignore[union-attr]
                 }
 
-            elif "latlon" in geo:
-                # ending up here means we either have the exact lat,lon format, or NaN
-                # skipping row if NaN
-                if row[geo["latlon"]] is None:
+            elif "latlon" in geo or "lonlat" in geo:
```
Pierlou, line 169: Keeping the insight
```suggestion
            elif "latlon" in geo or "lonlat" in geo:
                # skipping row if geo data is None
```

udata_hydra/analysis/geojson.py
```diff
@@ -29,6 +30,28 @@
 
 log = logging.getLogger("udata-hydra")
 
+
+# latlon/lonlat columns can contain values like "[48.8566, 2.3522]" or "48.8566 , 2.3522"
+# Both versions below strip spaces and brackets, then split on comma.
+
```
Pierlou: The insight is already given within the function, this feels misplaced here
```suggestion
```

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+    Returns (query, params) where params are the original column names used as
+    JSON keys via $1, $2… placeholders.
+    """
+    property_cols = [c for c in columns if c not in geo.values()]
+    params: list[str] = []
+    properties_fragments = []
+    for col in property_cols:
+        params.append(col)
+        placeholder = f"${len(params)}::text"
+        properties_fragments.append(f"{placeholder}, {_quote_ident(_db_col_name(col))}")
```
Pierlou: NIT
```suggestion
    for idx, col in enumerate(property_cols):
        params.append(col)
        properties_fragments.append(f"${idx + 1}::text, {_quote_ident(_db_col_name(col))}")
```

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+
+    Returns (query, params) where params are the original column names used as
+    JSON keys via $1, $2… placeholders.
+    """
+    property_cols = [c for c in columns if c not in geo.values()]
+    params: list[str] = []
+    properties_fragments = []
+    for col in property_cols:
+        params.append(col)
+        placeholder = f"${len(params)}::text"
```
Pierlou: Does that mean we're exporting all values as text? In which case it is not what we want (columns that contain ints/floats should be exported as such)

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
-    with output_file_path.open("w") as f:
-        json.dump(template, f, indent=4, ensure_ascii=False, default=str)
+            with output_file_path.open("w") as f:
+                f.write('{"type": "FeatureCollection", "features": [\n')
+                first = True
+                async for row in cursor:
+                    if not first:
+                        f.write(",\n")
+                    f.write(row[0])
+                    first = False
```
Pierlou, line 374: I don't think I understand the `first` trick: why don't we `f.write(f"{row[0]},\n")` for all rows? (in which case we should finally `f.write("]")`)

Pierlou (commented):
Thanks for the refactor 🙏 The flow looks good to me (though I haven't gone through the technicity of the SQL part), but I have a doubt regarding the type of the properties we export, and a couple of remarks.

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+    async with db.acquire() as conn:
+        async with conn.transaction():
+            cursor = conn.cursor(query, *params)
 
-    with output_file_path.open("w") as f:
-        json.dump(template, f, indent=4, ensure_ascii=False, default=str)
+            with output_file_path.open("w") as f:
+                f.write('{"type": "FeatureCollection", "features": [\n')
+                first = True
+                async for row in cursor:
```
bolinocroustibat, line 370: What happens in case of an error while parsing the rows?
Should we delete output_file_path in a try/finally if we detect failure after partial write, or wrap the stream loop in try/finally to remove the output file on any error after open?

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+
+
+async def csv_to_geojson_from_db(
+    table_name: str,
+    inspection: dict,
+    output_file_path: Path,
+    upload_to_minio: bool = True,
+) -> tuple[int, str | None] | None:
+    """Generate a GeoJSON file by streaming features directly from PostgreSQL."""
+    geo = _detect_geo_columns(inspection)
```
bolinocroustibat, line 354: NIT: a type hint would be nice here for readability:

`geo: dict | None = _detect_geo_columns(inspection)`

udata_hydra/analysis/geojson.py
bolinocroustibat: ... wouldn't `db_to_geojson` and `db_to_parquet` be even more explicit, and more consistent with namings like `csv_to_geojson`?

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
+    return query, params
+
+
+async def csv_to_geojson_from_db(
+    table_name: str,
+    inspection: dict,
+    output_file_path: Path,
+    upload_to_minio: bool = True,
+) -> tuple[int, str | None] | None:
+    """Generate a GeoJSON file by streaming features directly from PostgreSQL."""
```
bolinocroustibat: NIT: `csv_to_geojson_from_db` has a one-line docstring while `csv_to_geojson` documents args, behavior (skipped rows), and return values. Aligning the DB variant (even briefly: inputs, streaming, parity goal with the CSV path) might make it easier to maintain and choose between the two in case we have to.

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +249,123 @@ def get_features(
...
 
-    with output_file_path.open("w") as f:
-        json.dump(template, f, indent=4, ensure_ascii=False, default=str)
+            with output_file_path.open("w") as f:
+                f.write('{"type": "FeatureCollection", "features": [\n')
+                first = True
+                async for row in cursor:
+                    if not first:
+                        f.write(",\n")
+                    f.write(row[0])
```
bolinocroustibat: NIT: Might be a good opportunity to ad a column alias on the outer SELECT in `_build_feature_sql`, to avoid using `row[0]`, prefering somtehing like `row["feature_json"]`, that would massively improve readability imho.

udata_hydra/analysis/geojson.py
```diff
@@ -201,6 +206,35 @@ def get_features(
...
+    if geo is None:
+        log.debug("No geographical columns found, skipping")
+        return None
+
+    template = {"type": "FeatureCollection"}
+
+    template["features"] = streamable_list(get_features(file_path, inspection, geo))
+
+    with output_file_path.open("w") as f:
+        json.dump(template, f, indent=4, ensure_ascii=False, default=str)
```
bolinocroustibat, line 216: GeoJSON files are often large, pretty-printing can add a meaningful amount of redundant bytes on disk and for uploads, even if compression narrows the gap. 

Also `indent=4` gets a pretty-printed JSON while the database path stream does not re-indent the whole document like `indent=4` would.  Someone diffing two exports, grepping a file might think something broke when they see the compact stream.

Should we treat compact as the normal export format for both paths, dropping indent=4 on the CSV path? And assume if a human needs to read the file, s.he will format it automatically anyway?

bolinocroustibat (approved):
Thank you. A few remarks otherwise LGTM

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267

udata_hydra/analysis/geojson.py
ThibaudDauce: No here we talk about the columns/keys, I've added a comment in df12267

udata_hydra/analysis/geojson.py
ThibaudDauce: It will have a trailing comma in the JSON?

udata_hydra/analysis/geojson.py
ThibaudDauce: `_detect_geo_columns` already have a return-type? You're IDE should provide you the type hint on hover, no?

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267ae454d1a35cc6f67ed606ef3be2276231

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in df12267ae454d1a35cc6f67ed606ef3be2276231

udata_hydra/analysis/geojson.py
ThibaudDauce: Yes but this path is only for csv_to_geojson (the line only moved), I think we can leave it since it will be removed soon?

udata_hydra/analysis/geojson.py
ThibaudDauce: The caller `csv_to_geojson_and_pmtiles` in `csv.py:193-194` already wraps the call in a try/except that calls `remove_remainders(resource_id, ["geojson", "pmtiles", "pmtiles-journal"])`. So if an error occurs during streaming, the partial file is cleaned up by the caller.

This is the same pattern as the CSV path — `csv_to_geojson` doesn't handle cleanup on error either, it's delegated to the caller.

udata_hydra/analysis/geojson.py
bolinocroustibat: My IDE doesn't :) I type hint when not obvious at first sight

udata_hydra/analysis/geojson.py
```diff
@@ -452,7 +464,7 @@ async def csv_to_geojson_and_pmtiles(
 
     # Convert to GeoJSON — from DB if available and enabled, otherwise from CSV file
     if config.DB_TO_GEOJSON and table_name:
-        result = await csv_to_geojson_from_db(
+        result = await save_as_geojson_from_db(
```
bolinocroustibat: I think we should use this opportunity to rename:
`save_as_geojson_from_db` -> `db_to_geojson`
`save_as_parquet_from_db`-> `db_to_parquet`
Those names match the `source_to_target` style we already use (csv_to_parquet, csv_to_db, geojson_to_pmtiles, parquet_to_db), they also line up nicely with our config flags.

Optionally to match with that we could also rename `save_as_parquet` to `rows_to_parquet`

bolinocroustibat (approved):
Better naming suggestion

EDIT: doesn't pass the linting

udata_hydra/analysis/geojson.py
ThibaudDauce: Done in 2c31872

udata_hydra/analysis/geojson.py
bolinocroustibat: OK let's keep it like that

bolinocroustibat (approved):
LGTM, thank you!

udata_hydra/analysis/geojson.py
Pierlou: Indeed, in a parallel universe where JSON allows trailing commas the world is better 🌴

udata_hydra/analysis/geojson.py
```diff
@@ -215,26 +246,133 @@ def get_features(
...
-    template["features"] = streamable_list(get_features(file_path, inspection, geo))
+    db = await context.pool("csv")
+    async with db.acquire() as conn:
+        async with conn.transaction():
+            cursor = conn.cursor(query, *params)
 
-    with output_file_path.open("w") as f:
-        json.dump(template, f, indent=4, ensure_ascii=False, default=str)
+            with output_file_path.open("w") as f:
+                f.write('{"type": "FeatureCollection", "features": [\n')
```
Pierlou: So we have carriage returns but no indentation, is this what we want? It could also be a one-line JSON with space instead of `\n` (following [this remark](https://github.com/datagouv/hydra/pull/404/changes#r3094048303)) (doesn't really matter technically)

Pierlou (approved):
Nice, thanks for the refactoring 👏

udata_hydra/analysis/geojson.py
ThibaudDauce: I removed the newlines in [00711ba](https://github.com/datagouv/hydra/pull/404/commits/00711ba248683091ebabbcc2fd5dac425f78cab1)

Issue #405, opened 2026-04-03
Confusion between catalog and checks keys


Pierlou:
In the endpoints that get checks (`/api/checks/all` and `/api/checks/latest`), the result is ambiguous because of column names that are both in catalog and checks tables, for instance `url`, which then yields erroneous output in the API. See for instance [the checks of this resource](https://crawler.data.gouv.fr/api/checks/all/?resource_id=3f73d129-6b24-45cd-95e9-9bacc216d9d9):
```json
  {
    "id": 44640778,
    "catalog_id": 261555,
    "url": "https://cellar-c2.services.clever-cloud.com/ma-cantine-egalim-prod/media/open_data/registre_cantines.csv?v=202604031221",
    "domain": "cellar-c2.services.clever-cloud.com",
    "created_at": "2026-04-03T10:21:45.420703+00:00",
    "status": 200,
    ...
  },
  {
    "id": 44631349,
    "catalog_id": 261555,
    "url": "https://cellar-c2.services.clever-cloud.com/ma-cantine-egalim-prod/media/open_data/registre_cantines.csv?v=202604031221",
    "domain": "cellar-c2.services.clever-cloud.com",
    "created_at": "2026-04-03T04:26:25.176009+00:00",
    "status": 200,
    ...
  },
  ...
```
The remote URL of this resource changes a couple of times per day, but all the checks show that the checked URL is the same, because it's not the check URL we are displaying, but the current catalog URL

PR #406, opened 2026-04-03
Fix URL in checks

Pierlou:
Fixes https://github.com/datagouv/hydra/issues/405


bolinocroustibat (approved):
Indeed that was not explicit at all and quite confusing, thanks for spotting that!

PR #408, opened 2026-04-03
feat(csv): split GeoJSON and PMTiles into low-priority RQ jobs

bolinocroustibat:
Replaced by https://github.com/datagouv/hydra/pull/425.


udata_hydra/analysis/helpers.py
```diff
@@ -35,6 +42,22 @@ async def read_or_download_file(
...
+            resp.raise_for_status()
+            body = await resp.read()
+    fd, raw_path = tempfile.mkstemp(suffix=suffix)
+    path = Path(raw_path)
+    try:
+        os.write(fd, body)
+    finally:
+        os.close(fd)
+    log.debug(f"Downloaded {url} to {path} ({len(body)} bytes)")
+    return path
```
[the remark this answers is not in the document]
bolinocroustibat: It's possible, we are already doing it in `utils/files.py` -> `download_resource`.  We could use a shared function between the two.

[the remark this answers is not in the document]
bolinocroustibat: Refactored `utils/files.py` to use streaming HTTP for `download_url_to_tempfile`, also relocated and unify it with `download_resource`: https://github.com/datagouv/hydra/pull/408/changes/357dfb0b724cc81ae2c8d8ea12432acfadd416d9

udata_hydra/analysis/csv.py
```diff
@@ -181,24 +202,20 @@ async def analyse_csv(
...
-            remove_remainders(resource_id, ["geojson", "pmtiles", "pmtiles-journal"])
-            raise ParseException(
-                message=str(e),
-                step="geojson_export",
-                resource_id=resource_id,
-                url=url,
-                check_id=check["id"],
-            ) from e
+        if config.DB_TO_GEOJSON:
+            if not config.CSV_TO_DB:
```
Pierlou, line 206: I am confused with all the possible conversions: do we still have/want the option to do csv => geojson/parquet ? In which case there could still be two switches (do we want csv => `<other format>`, and do we allow to do it from the db)
Or will it now always be db => geojson/parquet?

udata_hydra/analysis/geojson.py
```diff
@@ -129,306 +125,78 @@ async def analyse_geojson(
         await Resource.update(resource_id, {"status": None})
 
 
-async def csv_to_geojson(
-    file_path: str,
+async def export_geojson_for_csv_analysis(
```
Pierlou, line 128: I find this name confusing, could we keep `db_to_geojson`? It's what this function does, right?

udata_hydra/analysis/geojson.py
```diff
@@ -129,306 +125,78 @@ async def analyse_geojson(
...
+    if check_id is not None:
+        await Check.update(
+            check_id,
+            {"pmtiles_url": pmtiles_url, "pmtiles_size": pmtiles_size},
         )
-    else:
-        pmtiles_url = None
-
+    do_unlink = (
+        config.REMOVE_GENERATED_FILES if unlink_geojson_after is None else unlink_geojson_after
```
Pierlou: NIT
```suggestion
        unlink_geojson_after or config.REMOVE_GENERATED_FILES
```

udata_hydra/analysis/resource.py
```diff
@@ -170,9 +170,10 @@ async def analyse_resource(
             await Resource.update(resource_id, data={"status": "TO_ANALYSE_CSV"})
             # Analyse CSV and create a table in the CSV database
             queue.enqueue(
-                analyse_csv,
+                task_analyse_csv,
```
Pierlou, line 173: Why do we need the proxy `task_analyse_csv`? Could we not keep the current syntax?
```suggestion
                analyse_csv,
```

udata_hydra/routes/status.py
```diff
@@ -189,7 +189,6 @@ async def get_health(request: web.Request) -> web.Response:
             "csv_to_parquet": config.CSV_TO_PARQUET,
             "db_to_parquet": config.DB_TO_PARQUET,
             "geojson_to_pmtiles": config.GEOJSON_TO_PMTILES,
-            "csv_to_geojson": config.CSV_TO_GEOJSON,
```
Pierlou, line 192: Following an upstream comment: maybe we should rationalize/change the syntax for these variables?

udata_hydra/utils/file.py
```diff
@@ -36,86 +38,129 @@ def extract_gzip(file_path: str) -> IO[bytes]:
...
             detected_extension = f".{match.group(1)}"
         else:
             detected_extension = ""
     else:
         # Not compressed - use magic to detect type
-        mime_type = magic.from_file(tmp_file.name, mime=True)
+        mime_type: str = magic.from_file(path, mime=True)
         detected_extension = mimetypes.guess_extension(mime_type) or ""
+        tmp_file = open(path, "rb")
+        tmp_file.close()
```
Pierlou, line 132: Why the open/close with nothing in between?

udata_hydra/utils/file.py
Pierlou, line 1: We are starting to have a bunch of download functions, aren't some of them redundant? Like `_http_get_to_temp_path` and `download_url_to_tempfile`?

udata_hydra/analysis/helpers.py
```diff
@@ -72,7 +75,7 @@ async def notify_udata(resource: Record | None, check: Record | dict | None) ->
     if config.GEOJSON_TO_PMTILES and check.get("pmtiles_url"):
         payload["document"]["analysis:parsing:pmtiles_url"] = check.get("pmtiles_url")
         payload["document"]["analysis:parsing:pmtiles_size"] = check.get("pmtiles_size")
-    if config.CSV_TO_GEOJSON and check.get("geojson_url"):
+    if config.DB_TO_GEOJSON and check.get("geojson_url"):
```
Pierlou, line 78: If we keep a double switch for conversions, we may need to adapt this

udata_hydra/analysis/csv.py
```diff
@@ -77,12 +78,32 @@
 minio_client = MinIOClient(bucket=config.MINIO_PARQUET_BUCKET, folder=config.MINIO_PARQUET_FOLDER)
 
 
+async def task_analyse_csv(
```
Pierlou, line 81: As mentionned where it's queued, I don't get why we need to wrap `analyse_csv` up

udata_hydra/analysis/geojson.py
```diff
@@ -451,55 +219,228 @@ async def csv_to_geojson_and_pmtiles(
...
-        result = await db_to_geojson(
-            table_name,
-            inspection,
-            geojson_filepath,
-            upload_to_minio=True,
-        )
-    else:
-        result = await csv_to_geojson(file_path, inspection, geojson_filepath, upload_to_minio=True)
-    if result is None:
+    geo_result = await export_geojson_for_csv_analysis(
```
Pierlou, line 224: If I understand correctly that means we never do csv=>geojson anymore? If it's always made from db, maybe we can drop the code for csv?

udata_hydra/analysis/csv.py
Pierlou, line 187: Do we want to have the parquet conversion as a task too? (in another PR maybe, as this one is quite heavy already)

Pierlou (commented):
Thanks for the work! As mentionned in the comments, maybe some parts could be rationalized, and I am also wondering about the `analysis/` folder: maybe we should have a `conversion/` folder too, so that we don't mix up parquet/geojson analysis and conversion from another format?

udata_hydra/analysis/geojson.py
bolinocroustibat: We already have a `db_to_geojson` method in `csv_geojson.py` file, which is the "low-level" method to convert the DB info to geojson.

This `export_geojson_for_csv_analysis` method  is the CSV-analysis pipeline wrapper and does more than converting to GeoJSON:
- bails out if there is no table_name
- updates the resource status (CONVERTING_TO_GEOJSON)
- calls db_to_geojson(...)
- records timer marks
- updates the check with geojson_url / geojson_size
So the name is meant to say “this is the GeoJSON export step for the CSV analysis flow”, while the actual DB streaming work is named `db_to_geojson`.

Name is indeed a bit ugly though, open to suggestions!

udata_hydra/analysis/resource.py
bolinocroustibat: `task_analyse_csv` wraps the existing `analyse_csv` method. I would keep it that way to differenciate with the reusable routine:
- task_ = worker/job contract — same idea as task_csv_to_geojson / task_geojson_to_pmtiles: “this is what we enqueue.”
- analyse_csv = reusable routine — call it from tests, CLI, or other code without queuing it

udata_hydra/routes/status.py
bolinocroustibat: Totally, but I would do that in another PR!

udata_hydra/analysis/csv.py
bolinocroustibat: As mentioned for `resource.py` where the method is called, we might still want to call `analyse_csv` as a routine, for example in tests or in the CLI, without enqueing it, I think I would keep it that way!

udata_hydra/analysis/geojson.py
bolinocroustibat: The code for the conversion from CSV to GeoJSON have already been dropped.

But GeoJSON is still a part of the CSV resource pipeline: after the CSV is analyzed and loaded into PostgreSQL parsing table with `csv_to_db`, we build a GeoJSON from that table with `db_to_geojson`.

`export_geojson_for_csv_analysis` is the orchestration for that pipeline: it checks DB_TO_GEOJSON, requires table_name, updates resource status, calls db_to_geojson, updates the check with geojson_url / geojson_size, etc. The name is about where it sits (CSV analysis flow for a resource), not about reading GeoJSON from raw CSV.

I agree the name is not ideal though, open to suggestions!

udata_hydra/analysis/csv.py
bolinocroustibat: As listed in this PR description in "Suggested in later, separate PRs", I think we should later completely drop csv_to_parquet logic later.

For the sake of readability of this already big PR, this was intentionally not part in this PR, but kept for later. `csv_to_geojson` has been dropped because it was more tightly tied to the new design.

Does that make sense?

udata_hydra/analysis/helpers.py
bolinocroustibat: `csv_to_geojson` logic has been dropped in this PR. It doesn’t work there unless you re-download or keep the file around, which would have been messy to keep in this PR.

Note: Parquet generation though, is still built in the same step as the CSV analysis, before the temp file is removed. So using the CSV file for Parquet is still fine today, so we keep it in this PR and I suggest we remove it in a later PR.
If we were instead to ever move csv_to_parquet to a separate job, we would hit the same problem as csv_to_geojson, though. Also, this PR is about splitting geojson and pmtiles generation into separate jobs, not parquet.

Does that make sense?

udata_hydra/utils/file.py
bolinocroustibat: This PR already refactors utils/files.py to use streaming HTTP for download_url_to_tempfile and relocate and unify it with download_resource: https://github.com/datagouv/hydra/commit/357dfb0b724cc81ae2c8d8ea12432acfadd416d9

It's already a download functions unifying step, do we want to do do more unifying in the same PR?

udata_hydra/analysis/csv.py
bolinocroustibat: In the future, I think we won't want to to CSV to parquet and CSV to GeoJSON anymore.

In this PR, we removed only the CSV to GeoJSON logic, because it is tightly linked to the logic of splitting GeoJSON and PMtiles generation as separate RQ jobs, as keeping it would need to re-download or keep the file around, which would have been messy to keep in this PR.

In a separate PR, when we'll tackle parquet, as suggested in the PR description, we should remove also the csv_to_parquet logic and put db_to_parquet as a separate queue.

Does that make sense?

udata_hydra/analysis/resource.py
Pierlou: In my mind there could be no difference between the function that does the work, and the task itself. I see it as: if we call the function, it does its stuff, and we can call it as a task too (similarly to Airflow tasks, that are just decorated but remain "technical" functions). If having a wrapper is a conceptually/programmatically better way, then we can stick to it of course

udata_hydra/utils/file.py
Pierlou: Maybe in a following one then, but I'm starting to get lost 😅

udata_hydra/analysis/geojson.py
Pierlou: `csv_to_geojson` still exists [here](https://github.com/datagouv/hydra/blob/286eba52ad8d0044baf3ce1c835387700e519cef/udata_hydra/analysis/csv_geojson.py#L41) and is still used in the CLI and in the tests. Do we want to keep it?

udata_hydra/analysis/resource.py
bolinocroustibat: But when we call it from the tests and/or from the CLI, that wouldn't be an issue you think?

PR #414, opened 2026-04-10
fix: use a default max size for unknown format

maudetes:
Improve format literal typing and add a default max size.

Return an explicit "unknown" literal if format couldn't be detected instead of a fallback on "csv" or ""

Fix https://errors.data.gouv.fr/organizations/sentry/issues/284888/


udata_hydra/config_default.toml
```diff
@@ -66,6 +66,8 @@ MAX_FILESIZE_ALLOWED.parquet = 52428800
 MAX_FILESIZE_ALLOWED.wms = 1048576     # 1 MB
 MAX_FILESIZE_ALLOWED.wfs = 1048576     # 1 MB
 
+DEFAULT_MAX_FILESIZE_ALLOWED = 52428800
```
Pierlou, line 69: Is this arbitrary or based on something?

Pierlou (approved):
Thanks for the fix 🙏

udata_hydra/config_default.toml
maudetes: Today we used to fallback on csv (or fail) with a current default value of 100Mo.
It made sense to me to have a filesize limit lowered (divided by two) for unknown formats compared to the maximum?
In any case, it's a default setting value, we can override it easily if needed.

PR #415, opened 2026-04-15
feat(s3): S3 Object Storage via boto3

### LINKED ISSUE #2000: [hydra] Migration du stockage sur OVH Object Storage
Pour les pmtiles & geojson & parquet.

- [ ] Changer les configs côté infra
- [ ] Copie de minio vers OVH
- [ ] Migrer les URLs actuelles dans les extras

bolinocroustibat:
Closes https://github.com/datagouv/data.gouv.fr/issues/2000

Following the move from MinIO to S3-compatible Object Storage, switching MinIO Python SDK to a boto3-based stack.

This introduces `S3Client` with the same upload contract and public URL shape as before, swaps analysis callers to it, replaces the `minio` dependency with `boto3`, and updates tests to mock boto3.


udata_hydra/__init__.py
```diff
@@ -28,6 +28,22 @@ def configure(self) -> None:
             with open(Path(local_settings), "rb") as f:
                 configuration.update(tomllib.load(f))
 
+        _s3_env_keys = (
```
Pierlou: Why do we need this now and not previously? 🤔

Pierlou (approved):
Thanks for the migration 🙏

udata_hydra/__init__.py
bolinocroustibat: Not needed, that was a fallback, I removed it.

udata_hydra/utils/s3.py
```diff
@@ -0,0 +1,66 @@
...
+class S3Client:
+    def __init__(self, bucket: str, prefix: str | None = None):
+        self.user = config.S3_ACCESS_KEY_ID
+        self.password = config.S3_SECRET_ACCESS_KEY
+        self.bucket = bucket
+        self.prefix = _normalize_prefix(prefix)
+        host = config.S3_ENDPOINT or "test"
+        self._resource = boto3.resource(
+            "s3",
+            endpoint_url=f"https://{host}",
```
Pierlou: NIT
```suggestion
        self._resource = boto3.resource(
            "s3",
            endpoint_url=f"https://{config.S3_ENDPOINT or 'test'}",
```

udata_hydra/utils/s3.py
```diff
@@ -0,0 +1,66 @@
...
+        path = Path(file_path)
+        if path.is_file():
+            file_name = path.name
+            key = f"{self.prefix}/{file_name}" if self.prefix else file_name
+            self._resource.Bucket(self.bucket).upload_file(str(path), key)
+            if delete_source:
+                path.unlink()
+            if self.prefix:
+                return f"https://{config.S3_ENDPOINT}/{self.bucket}/{self.prefix}/{file_name}"
+            return f"https://{config.S3_ENDPOINT}/{self.bucket}/{file_name}"
```
Pierlou: These won't work for all S3 hosts:
- with Minio this is the correct syntax (`https://{endpoint_url}/{bucket}/{object_key}`)
- but with OVH for instance the syntax is `https://{bucket}.{endpoint_url}/{object_key}`

So either we have something most sophisticated, or we create an issue to remember to change this when we change host

Pierlou (approved):
Thanks for the migration 🙏  just concerned about the public URL syntax

udata_hydra/utils/s3.py
bolinocroustibat: Quite fundamental indeed. We can change it on this branch, there is no backward compatibility with MinIO in this branch. Changed [here](https://github.com/datagouv/hydra/pull/415/changes/54c59ab42f0bfdbd12b6007221a85322d38e5842)

PR #416, opened 2026-04-16
fix: use basename as param in jobs 

maudetes:
Indeed, `TEMPORARY_DOWNLOAD_FOLDER` may differ between workers.

Fix https://errors.data.gouv.fr/organizations/sentry/issues/285991/

> [!IMPORTANT]
> this should also apply when splitting analysis among different jobs in https://github.com/datagouv/hydra/pull/408 if they require a filename to be passed as param


Pierlou (approved):
Thanks for the fix 🙏

PR #422, opened 2026-05-12
refactor(conversion): separate conversion methods into dedicated files in `/conversion`

bolinocroustibat:
### `udata_hydra/conversion/schema.py` (new)
- `PYTHON_TYPE_TO_PG`, `PYARROW_TYPE_TO_PYTHON`, `PYTHON_TYPE_TO_PA` — from `analysis/csv.py`, `analysis/parquet.py`, and `utils/parquet.py` respectively
- `compute_create_table_query` — from `analysis/csv.py`

### `udata_hydra/conversion/csv_to_db.py` (new)
- CSV → PostgreSQL import and table creation — from `analysis/csv.py`

### `udata_hydra/conversion/csv_to_parquet.py` (new)
- `tabular_rows_to_parquet`, `csv_to_parquet` — from `utils/parquet.py` (and the CSV analysis flow in `analysis/csv.py`)

### `udata_hydra/conversion/db_to_parquet.py` (new)
- `db_to_parquet` — from `utils/parquet.py`

### `udata_hydra/conversion/parquet_to_db.py` (new)
- Parquet file → DB pipeline — from `analysis/parquet.py`

### `udata_hydra/conversion/csv_to_geojson.py`, `db_to_geojson.py`, `geojson_to_pmtiles.py`, `csv_to_geojson_and_pmtiles.py` (new)
- GeoJSON and PMTiles export paths — from `analysis/geojson.py` (and related CSV-side orchestration)

### `udata_hydra/conversion/__init__.py` (new)
- CSV `field_size_limit` bootstrap — from `analysis/csv.py`

### `udata_hydra/analysis/tables_index.py` (new)
- `get_previous_analysis`, `insert_tables_index_entry` — logic from `analysis/csv.py` / `analysis/parquet.py` (metadata row in `tables_index`, not SQL indexes)

### `udata_hydra/utils/parquet.py`
- Keeps only `detect_parquet_from_headers` — Parquet write/conversion logic moved to `conversion/`

### `udata_hydra/cli.py`
- `csv_to_geojson`, `geojson_to_pmtiles` imports — now from `conversion/` instead of `analysis/geojson.py`

### Tests
- Imports and `patch` targets updated to match the new modules; session config sets `TEMPORARY_DOWNLOAD_FOLDER=/tmp` so queued analysis basename lookups match downloads under pytest.

## Renames (disambiguation)

[...]

PR #423, opened 2026-05-12
feat(analysis): GeoJSON export from DB only in CSV pipeline

bolinocroustibat:
Needs [#422 (`refactor/conversion-package`)](https://github.com/datagouv/hydra/pull/422) to be merged first.

Remove csv_to_geojson logic entirely in the analysis/conversion pipeline, only relying on db_to_geojson anymore for GeoJSON file generation.


udata_hydra/analysis/csv.py
```diff
@@ -137,14 +137,14 @@ async def analyse_csv(
...
-                table_name=table_name if config.CSV_TO_DB else None,
-            )
+            if config.DB_TO_GEOJSON:
+                await db_to_geojson_and_pmtiles(
+                    table_name=table_name,
+                    inspection=csv_inspection,
+                    resource_id=resource_id,
+                    check_id=check["id"],
+                    timer=timer,
+                )
```
Pierlou: The condition is checked at the beginning of `db_to_geojson_and_pmtiles`
```suggestion
            await db_to_geojson_and_pmtiles(
                table_name=table_name,
                inspection=csv_inspection,
                resource_id=resource_id,
                check_id=check["id"],
                timer=timer,
            )
```

udata_hydra/conversion/csv_to_db.py
```diff
@@ -31,10 +31,6 @@ async def csv_to_db(
     :table_name: used to create tables
     :debug_insert: insert record one by one instead of using postgresql COPY
     """
-    if not config.CSV_TO_DB:
```
Pierlou, line 34: Do we want to kill this switch? We will tend to have most conversions relying of the db, but we could want to keep this one as the main switch?

udata_hydra/analysis/csv.py
```diff
@@ -137,14 +137,14 @@ async def analyse_csv(
...
         try:
-            await csv_to_geojson_and_pmtiles(
-                file_path=tmp_file.name,
-                inspection=csv_inspection,
-                resource_id=resource_id,
-                check_id=check["id"],
-                timer=timer,
-                table_name=table_name if config.CSV_TO_DB else None,
-            )
+            if config.DB_TO_GEOJSON:
```
ThibaudDauce: Already done inside the function?

ThibaudDauce (approved):

udata_hydra/config_default.toml
```diff
@@ -93,23 +92,22 @@ MINIO_USER = ""
 MINIO_PWD = ""
 
 # -- Parquet conversion settings -- #
-CSV_TO_PARQUET = false
+CSV_TO_PARQUET = false # to deprecated
```
Pierlou, line 96: Can't we delete it?

udata_hydra/config_default.toml
```diff
@@ -93,23 +92,22 @@ MINIO_USER = ""
...
-MINIO_PMTILES_BUCKET = ""
-MINIO_PMTILES_FOLDER = "" # no trailing slash
-
 # -- Geojson conversion settings -- #
-CSV_TO_GEOJSON = false
 DB_TO_GEOJSON = false
 MINIO_GEOJSON_BUCKET = ""
 MINIO_GEOJSON_FOLDER = "" # no trailing slash
 
+# -- PMTiles conversion settings -- #
```
Pierlou, line 107: Could we leave the block where it was to prevent a change?

pyproject.toml
```diff
@@ -75,7 +75,7 @@ markers = [
     "slow: marks tests as slow (deselect with '-m \"not slow\"')",
     "catalog_harvested: use catalog_harvested.csv as source",
 ]
-addopts = "--cov=udata_hydra --cov-report term-missing --cov-branch --cov-fail-under=80"
+addopts = "--cov=udata_hydra --cov-report term-missing --cov-branch --cov-fail-under=79"
```
Pierlou, line 78: Why? 👀

Pierlou (commented):
Thanks for the cleanup!
Do we want to keep the branching for csv/db to geojson [here](https://github.com/datagouv/hydra/blob/3a1a7d1bb72d02fbf6bc0c29250f81493e99938f/udata_hydra/conversion/csv_to_geojson_and_pmtiles.py#L44)?

udata_hydra/config_default.toml
bolinocroustibat: Yes but in a separate PR: https://github.com/datagouv/hydra/pull/424

udata_hydra/config_default.toml
bolinocroustibat: Having the block geojson before the block pmtiles to follow the pipeline order improves readability and it's the right opportunity PR to change that no?

pyproject.toml
bolinocroustibat: We lost 1% coverage so it's for tests to pass... Will improve the tests coverage in a later PR.

udata_hydra/config_default.toml
Pierlou: Fair enough

udata_hydra/conversion/csv_to_db.py
bolinocroustibat: Yes, I agree, let's keep it, can be useful for debugging on preprod/prod, etc.
Done here: https://github.com/datagouv/hydra/pull/423/changes/dc59285dd122c74685aa383bca781d163fcd7b7c (the switch has been slightly moved to be around the calls and check update instead, though)

Pierlou: Don't we need to have it on top of the other `DB_TO_xxx` switches everywhere too? If the db insertion has not been done, we won't be able to perform the downstream conversions from the table

udata_hydra/analysis/csv.py
bolinocroustibat: Indeed, well spotted. For better readability and maintenability I would suggest to have all those switches directly in the pipeline outside the functions so we can them see clearly in the pipeline.
Done here: https://github.com/datagouv/hydra/pull/423/changes/3b6b1125d50810ce23b3ee7d71b18ae5669700ed

Pierlou: Thanks, maybe we should do the same for all similar cases (in an upcoming PR I guess)

bolinocroustibat: It has also been done for `CSV_TO_DB` in this PR I believe.

For `CSV_TO_PARQUET`, it will be removed in https://github.com/datagouv/hydra/pull/424.

Regarding `GEOJSON_TO_PMTILES` and `PARQUET_TO_DB`, it's a bit more tricky as it currently deactivates the whole geojson analysis (`analyse_geojson`) /  parquet analysis (`analyse_parquet`). We could:
1) Leave it as it is for now
2) Rename those two keys  `GEOJSON_TO_PMTILES` -> `GEOJSON_ANALYSIS_ENABLED` / `PARQUET_TO_DB` -> `PARQUET_ANALYSIS_ENABLED` (similar to `OGC_ANALYSIS_ENABLED`)
3) add new switches for analysis, and move those switches later in the analysis pipelines so that they don't turn off the analysis, but only the conversion - but I would prefer not having switches that depend on each other

Pierlou: Maybe we leave it for now (this PR is quite heavy already) and have a dedicated PR for unification/renaming?

udata_hydra/analysis/csv.py
bolinocroustibat: Indeed, see https://github.com/datagouv/hydra/pull/423#discussion_r3233128022

udata_hydra/conversion/csv_to_db.py
bolinocroustibat: Sorry indeed: https://github.com/datagouv/hydra/pull/423/changes/18c6a7a45047dc58c13d2241412bfdb4c05f443c

PR #424, opened 2026-05-12
feat: db only parquet export

bolinocroustibat:
Needs https://github.com/datagouv/hydra/pull/422 and https://github.com/datagouv/hydra/pull/423 to be merged first.


tests/test_analysis/test_analysis_csv.py
```diff
@@ -912,6 +912,50 @@ async def test_too_long_column_name(
...
+):
+    """export_db_to_parquet runs only when DB_TO_PARQUET is enabled (analyse_csv path)."""
+    check = await fake_check()
+    url = check["url"]
+    table_name = hashlib.md5(url.encode("utf-8")).hexdigest()
+    rmock.get(url, status=200, body=catalog_content)
+
+    with (
+        patch("udata_hydra.config.DB_TO_PARQUET", db_to_parquet_enabled),
+        patch("udata_hydra.config.MIN_LINES_FOR_PARQUET", 1),
```
ThibaudDauce: I think `MIN_LINES_FOR_PARQUET` is only used inside `export_db_to_parquet` so not sure it's useful to mock it?

ThibaudDauce (approved):

tests/test_analysis/test_analysis_csv.py
bolinocroustibat: You're right, since we are mocking `export_db_to_parquet`, we don't need this. Removed [here](https://github.com/datagouv/hydra/pull/424/changes/d9cbe5ea872691a5d4168608857e08e85502f28b)

PR #425, opened 2026-05-12
feat(csv): parquet and GeoJSON exports on dedicated RQ queues

### LINKED ISSUE #412: [hydra] Découper les tâches d'analyse csv en sous-tâches
La tâche [`analyse_csv`](https://github.com/datagouv/hydra/blob/a72fa350fd37799eaf8c8e1682c03cf6fe5dd50d/udata_hydra/analysis/csv.py#L78) fait aujourd'hui le `csv_to_db`, `csv_to_parquet` et le `csv_to_geojson_and_pmtiles` le cas échéant.

Depuis que le dataframe n'est plus chargé en mémoire (cf https://github.com/datagouv/hydra/issues/314) , il pourrait être intéressant de séparer les tâches pour pouvoir les pousser en queue (et ainsi avoir un niveau de priorité plus précis (ex différencier mise en db csv VS la conversion aux formats géographiques en `low`)).

Voir aussi https://github.com/datagouv/hydra/pull/402 depuis la DB directement plutôt que depuis le fichier source.

bolinocroustibat:
Closes https://github.com/datagouv/hydra/issues/412 (and duplicate https://github.com/datagouv/data.gouv.fr/issues/1991).

Needs https://github.com/datagouv/hydra/pull/422, https://github.com/datagouv/hydra/pull/423 and https://github.com/datagouv/hydra/pull/424 to be merged first.

Replaces https://github.com/datagouv/hydra/pull/408.

Parquet and GeoJSON/PMTiles coversion steps were bound to the same RQ job as CSV ingest, so they competed for the same worker capacity and could not be scaled or deprioritized independently.

After CSV ingest finishes (`analyse_csv`), exports are enqueued as separate low-priority RQ jobs (`export_parquet`, `export_geojson_pmtiles`). Thos jobs run on the **`low`** queue without blocking ingest on **`default`** / **`high`**.

Note: GeoJSON and PMTiles stay in one job in the same process so tippecanoe keeps using a local file.


udata_hydra/analysis/exports.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+from udata_hydra.analysis import helpers
+from udata_hydra.conversion.db_to_geojson_and_pmtiles import db_to_geojson_and_pmtiles
+from udata_hydra.db.check import Check
+from udata_hydra.db.resource import Resource
+from udata_hydra.utils import ParseException, handle_parse_exception, remove_remainders
+
+
+async def export_parquet(
```
ThibaudDauce: `export_parquet` and `export_geojson_pmtiles` are really similar, could we refactor them with an helper?

udata_hydra/analysis/csv.py
```diff
@@ -217,10 +209,15 @@ async def export_db_to_parquet(
     if resource_id:
         await Resource.update(resource_id, {"status": "CONVERTING_TO_PARQUET"})
 
+    out_base = (
+        str(Path(config.TEMPORARY_DOWNLOAD_FOLDER or "/tmp") / resource_id)
```
ThibaudDauce: I think `remove_remainders` use the current directory for cleanup and not `TEMPORARY_DOWNLOAD_FOLDER`

bolinocroustibat: God catch, fixed.
But by the way, isn't that more stable to use TEMPORARY_DOWNLOAD_FOLDER back everywhere for big files? Writing large temp files under an unpredictable cwd could be fragile, a configured temp folder is safer no?

udata_hydra/analysis/exports.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/425/changes/e23ece4bab71bfe7556446c7be3e198bff76ee49

udata_hydra/analysis/csv.py
ThibaudDauce: I think it's better but can be done in another PR?

ThibaudDauce (approved):

README.md
```diff
@@ -86,6 +86,8 @@ To empty all the queues:
 
 `uv run rq empty -c udata_hydra.worker low default high`
 
+CSV Parquet and GeoJSON/PMTiles exports from parsed tables are queued as separate jobs on the **`low`** queue (`export_parquet`, `export_geojson_pmtiles`), so you can run dedicated workers or tune capacity for heavy MinIO and tippecanoe work without blocking CSV ingest on **`default`** / **`high`**.
```
Pierlou: Preshot 👀 
```suggestion
CSV Parquet and GeoJSON/PMTiles exports from parsed tables are queued as separate jobs on the **`low`** queue (`export_parquet`, `export_geojson_pmtiles`), so you can run dedicated workers or tune capacity for heavy S3 and tippecanoe work without blocking CSV ingest on **`default`** / **`high`**.
```

udata_hydra/analysis/helpers.py
```diff
@@ -64,7 +64,7 @@ async def notify_udata(resource: Record | None, check: Record | dict | None) ->
             else None,
         },
     }
-    if config.CSV_TO_DB and check.get("parsing_table"):
+    if check.get("parsing_table"):
```
Pierlou: We drop it here but keep it in `udata_hydra/analysis/csv.py`? Maybe it's time for a big cleanup of the config (default and on infra) 🧹

udata_hydra/analysis/exports.py
```diff
@@ -0,0 +1,89 @@
...
+    """RQ target: GeoJSON + PMTiles export for a parsed CSV resource."""
+    await _run_export_job(
+        table_name=table_name,
+        inspection=inspection,
+        resource_id=resource_id,
+        check_id=check_id,
+        url=url,
+        step="geojson_export",
+        remainder_types=["geojson", "pmtiles", "pmtiles-journal"],
+        export_fn=db_to_geojson_and_pmtiles,
```
Pierlou, line 88: I am wondering if we want to push it to enqueue geojson and then enqueue PMTiles, but that would require to manage the output geojson file (the conversion to PMTiles is not made from db), we can assess this later

Pierlou (approved):
Nice, thanks for the refactor 👏 cool idea to create the exports.py file

udata_hydra/analysis/exports.py
bolinocroustibat: That was the case in a first attempt, but it made the logic complicated and fragile so I decided to have a single queue for geosjon and pmtiles in this PR. We can think about it again in a later PR indeed.

udata_hydra/analysis/helpers.py
bolinocroustibat: You're right, we should keep it here for now

bolinocroustibat: https://github.com/datagouv/hydra/commit/04b3ac6ff6cad033986206dec2d395c40bd0f6a7

PR #429, opened 2026-05-26
One bucket per env, prefixes made from file extentions

Pierlou:


udata_hydra/utils/s3.py
```diff
@@ -59,12 +50,13 @@ def send_file(
         path = Path(file_path)
         if path.is_file():
             file_name = path.name
-            object_key = f"{self.prefix}/{file_name}" if self.prefix else file_name
+            file_ext = file_name.split(".")[-1]
```
bolinocroustibat: Would be even better/safer to use the nice Path object method:

`file_ext = path.suffix`

no?

udata_hydra/utils/s3.py
```diff
@@ -59,12 +50,13 @@ def send_file(
...
+            object_key = f"{file_ext}/{file_name}"
             self._resource.Bucket(self.bucket).upload_file(
                 str(path),
                 object_key,
                 ExtraArgs={
-                    "ContentType": CONTENT_TYPES[file_name.split(".")[-1]],
+                    "ContentType": CONTENT_TYPES[file_ext],
```
bolinocroustibat: `"ContentType": CONTENT_TYPES[path.suffix],`

bolinocroustibat (commented):
A little optimization otherwise LGTM

PR #430, opened 2026-05-29
Introduce data formats

Pierlou:
The current state of the repo inheritates from iterative additions (conversions to parquet, geojson, pmtiles...) which all share a relatively similar structure (download, analyse if relevant, convert, upload, notify). This means quite a lot of copied-pasted code everywhere and an extensive codebase. This refactor aims at rationalizing this, with the creation of the `DataFormat` class that describes how we always manipulate the same data, just in different formats, converted from eachother.

Main refactors anticipated:
- store each format's attributes and methods (mime-types, analysis, conversion functions...) at a unique place (no more `/utils/parquet-csv-geojson.py`, conversions within each folder)
- manipulate high level objects like:
```python
csv_file = Csv(path="file.csv", resource_id="...")
await csv_file.inspect()  # run csv-detective
table: Table = await csv_file.to_db(...)
geojson_file: Geojson = await table.to_geojson(...)
pmtiles_file: PMTiles = await geojson_file.to_pmtiles(...)
```
- rationalize the structure for identical processes (analysis, conversion, send to S3, notify udata) with the decorator-like function `_run_export_job`

We can group `Parquet` with `CsvLike` as a `Tabular` class when [upgrading csv-detective](https://github.com/datagouv/csv-detective/pull/253)


udata_hydra/data_formats/csv_like/__init__.py
```diff
@@ -0,0 +1,107 @@
...
+class Csvgz(CsvLike):
+    standard_mime_type = "application/gzip"
+    valid_mime_types = {
+        standard_mime_type,
+        "application/octet-stream",
+        "application/x-gzip",
+    }
+    max_filesize_allowed = int(config.MAX_FILESIZE_ALLOWED["csvgz"])
+    check_url = "csv.gz"
+
```
bolinocroustibat, line 202: Would it be a good idea to have a compress/decompress methods in that class?

udata_hydra/db/resource.py
```diff
@@ -27,8 +30,10 @@ class Resource:
         "CONVERTING_TO_GEOJSON": "csv is currently being converted to geojson",
         "TO_ANALYSE_PARQUET": "parquet resource content to be analysed",
         "ANALYSING_PARQUET": "retrieving parquet column metadata",
-        "TO_ANALYSE_OGC": "OGC service to be analysed",
-        "ANALYSING_OGC": "retrieving OGC service metadata",
+        "TO_ANALYSE_WFS": "WFS service to be analysed",
```
bolinocroustibat, line 33: Good catch, but this is not really related to the PR no? Should we do that in a separate PR, right away?

tests/conftest.py
```diff
@@ -250,13 +250,13 @@ async def db():
 
 @pytest_asyncio.fixture
 async def insert_fake_resource():
-    async def _insert_fake_resource(database, status: str | None = None):
```
bolinocroustibat, line 253: Well spotted, we could actually remove that in a quick separate PR!

tests/test_utils/test_s3_client.py
```diff
@@ -24,28 +25,34 @@ def mock_s3(mocker: MockerFixture) -> Iterator[MagicMock]:
...
 )
 def test_s3_client_upload(
-    mock_s3: MagicMock, tmp_path: Path, extension: str, patched_config: tuple[str, str], mocker
+    mock_s3: MagicMock,
+    tmp_path: Path,
+    data_format: DataFormat,
+    patched_config: tuple[str, str],
+    mocker,
 ) -> None:
+    extension = data_format.__class__.__name__.lower()
```
bolinocroustibat, line 49: Does that work with `CsvLike`? Maybe a `Class.EXTENSION` constant would be safer in case of future classes? Just a thought

udata_hydra/data_formats/geojson/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Geojson(DataFormat):
+    standard_mime_type = "application/vnd.geo+json"
+    valid_mime_types = {standard_mime_type}
+    max_filesize_allowed = int(config.MAX_FILESIZE_ALLOWED["geojson"])
+    check_url = "geojson"
+    further_analysis = True
+
+    async def analyse(self, check: dict):
```
bolinocroustibat: We could put all `analyse_xxxx` classes as method classes instead, which would make the classes long to read, but which would avoid this

udata_hydra/analysis/helpers.py
```diff
@@ -40,8 +41,10 @@ async def read_or_download_file(
             headers=json.loads(check.get("headers") or "{}"),
             max_size_allowed=None
             if exception
-            else int(
-                config.MAX_FILESIZE_ALLOWED.get(file_format, config.DEFAULT_MAX_FILESIZE_ALLOWED)
+            else (
```
bolinocroustibat, line 44: NIT: Is that compatible with the function return type (`IO[bytes]`) ? 
Do that work when type checked?

bolinocroustibat (commented):
That looks more readable and maintainable, I'm in!
A few comments and questions, might add more later

udata_hydra/data_formats/csv_like/__init__.py
Pierlou: It's currently in the utils but I agree that it would make a lot of sense to have it here, we can do that in a short following PR 🚀

udata_hydra/db/resource.py
Pierlou: It is, now the status we push before enqueuing is `f"TO_ANALYSE_{data_format.__name__.upper()}"` so I reflected this in the statuses descriptions. Does that sound fair? (though I kept `ANALYSING_CSV` and `VALIDATING_CSV` to prevent too much extensivity)

tests/test_utils/test_s3_client.py
Pierlou: There should be no instanciation of `CsvLike` ever in the code, so we should be fine here

udata_hydra/data_formats/geojson/__init__.py
Pierlou: Yes totally, I'll do that in an upcoming commit

udata_hydra/analysis/helpers.py
Pierlou: I believe so, we just move where the max_file_allowed is pulled from the config but the mechanism doesn't change

bolinocroustibat (approved):
Except for my few remarks, LGTM!

udata_hydra/analysis/exports.py
```diff
@@ -5,85 +5,123 @@
...
     remainder_types: list[str],
-    export_fn,
-) -> None:
-    check_out = None
+    export_fn: str,
+    upload_to_s3: bool = True,
+    delete_output: bool = False,
+    delete_input: bool = True,
+) -> DataFormat | None:
+    output, check_out = None, None
```
bolinocroustibat, line 30: I got some type errors here: `Type "None" is not assignable to declared type "DataFormat"
  "None" is not assignable to "DataFormat"``

udata_hydra/analysis/exports.py
```diff
@@ -5,85 +5,123 @@
...
-        resource = await Resource.get(resource_id)
-        if resource is not None and check_out is not None:
-            await helpers.notify_udata(resource, check_out)
-        await Resource.update(resource_id, {"status": None})
+        if data_object.resource_id is not None:
+            resource = await Resource.get(data_object.resource_id)
+            if resource is not None and check_out is not None:
+                await helpers.notify_udata(resource, check_out)
+            await Resource.update(data_object.resource_id, {"status": None})
+        return output
```
bolinocroustibat: I got an error regarding this: A "return" cannot be used to exit a "finally" block

udata_hydra/analysis/exports.py
```diff
@@ -5,85 +5,123 @@
...
+    delete_input: bool = True,
+) -> DataFormat | None:
+    output, check_out = None, None
     try:
-        await export_fn(table_name, inspection, resource_id, check_id)
+        output: "DataFormat" = await getattr(data_object, export_fn)()
+        if upload_to_s3:
+            df_name = output.__class__.__name__
+            log.debug(f"Uploading {df_name} file {output.path} to S3")
+            upload_url = s3_client.send_file(
```
bolinocroustibat: Some type errors here:

```
Argument of type "Unknown | Literal[False] | None" cannot be assigned to parameter "delete_source" of type "bool" in function "send_file"
  Type "Unknown | Literal[False] | None" is not assignable to type "bool"
    "None" is not assignable to "bool"
```

bolinocroustibat (changes requested):
While merging from main and solving the conflicts, I noticed a few type errors.
Nothing blocking but that might hep to notice slight inconsistencies.

bolinocroustibat (approved):
LGTM!

PR #431, opened 2026-06-01
Add `header` field to parquet analysis

Pierlou:
Explore expects the `header` field


maudetes (commented):
Thank you for the fix!
I think we'll need additional infos?
In [this code](https://github.com/datagouv/explore.data.gouv.fr/blob/main/src/apps/tabular/store/storeTabular.js#L94-L110), we seem to need:
- profile.header
- profile.profile[key]
- profile.columns

maudetes (approved):
Let's start with this

PR #432, opened 2026-06-02
fix(cli): apply --quiet to all loggers globally

bolinocroustibat:
- Use `logging.disable()` via a new `quiet_logs()` helper in the logger module, so all third-party loggers are silenced when using `--quiet` on CLI commands and logging is restored after purge and catalog commands.

- Expose `--quiet` as a Typer CLI arg on `purge-selected-csv-tables` for consistency across CLI commands.

### DIFF
--- a/tests/test_logger.py
+++ b/tests/test_logger.py
@@ -1,0 +1,38 @@
+import logging
+
+from udata_hydra.logger import quiet_logs
+
+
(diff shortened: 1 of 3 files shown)


Pierlou (approved):
Looks good to me, thanks for the fix! Are there any other functions that would require this new syntax?

PR #433, opened 2026-06-02
feat(status): include python version in health endpoint

bolinocroustibat:
Also improve unit testing of the keys/values of the health check endpoint


Pierlou (approved):
Clean ✨

PR #436, opened 2026-06-04
fix(cli): silence botocore logs when using --quiet

bolinocroustibat:
Related to [this thread](https://matrix.to/#/!FMXpGlfBgjvMumrnPQ:agent.dinum.tchap.gouv.fr/$zztxrWhri1K1po27nvGaosHmt8d9ShdxDmwqSEKqk1o?via=agent.dinum.tchap.gouv.fr) 

This PR ensures that:
- boto3/botocore is not initialized on every CLI startup when S3 is not needed
- startup and runtime stay quiet when the command is run with `--quiet`

When running a Hydra CLI command with `--quiet` arg , `botocore` still logged at startup because boto3 was initialized when Python imported `cli.py` (four S3Client(...) at module level), before `quiet_logs()` ran inside the command.

A `quiet` flag on `S3Client` would not help: imports happen before Typer reads `--quiet`, and that would mix CLI concerns into the S3 helper.

We add a small `cli_entry` that calls `logging.disable(WARNING) `when `--quiet` is in argv, then imports cli (same reason `run()` cannot live only at the bottom of `cli.py`: the module is already loaded). We also use `get_s3_client()` so `boto3` starts only on first upload; `reset_s3_client()` is for tests only.

Warning: CLI entrypoint has changed to `udata_hydra.cli_entry:run`

**To be done later:**

A follow-up PR could turn bloated `cli.py` into a `cli/` package, split commands across modules, and move `run()` (with early quiet) into `cli/__init__.py`, then drop `cli_entry.py`.


udata_hydra/cli_entry.py
```diff
@@ -0,0 +1,13 @@
...
+
+import logging
+import sys
+
+
+def run() -> None:
+    """CLI entry point: silence logs below ERROR when --quiet is in argv, before importing cli."""
+    if "--quiet" in sys.argv:
+        logging.disable(logging.WARNING)
+    from udata_hydra.cli import cli
```
Pierlou: This looks like a circular import but I'm guessing it's not, twisted my mind a bit

udata_hydra/utils/s3.py
Pierlou, line 1: Why this over a simple `create_client()` function where the S3Client would be instanciated? Or if needed, an `__open__` method in the client, to be able to do `with S3Client(...) as client: ...`

Pierlou (commented):
I am curious about the chosen syntax (with a global variable) over other possibiities, not that I like any over another, just genuinely intested in this choice 🤓

udata_hydra/cli_entry.py
bolinocroustibat: The import is inside run() so that logging.disable takes effect before udata_hydra.cli is first loaded, not after.

bolinocroustibat: Also, if that makes things more clear, in this later PR (https://github.com/datagouv/hydra/pull/437), we can avoid this new `cli_entry.py` file by creating a `/cli` module and putting the run without imports in `/cli/__init__.py` (and splitting the bloated `cli.py`).
I prefer doing this in those two separated PRs for clarity: this one with the `run()` call without the imports, then https://github.com/datagouv/hydra/pull/437 for the refactoring of cli.py.
Let me know if that makes sense

udata_hydra/utils/s3.py
bolinocroustibat: We want to call it only from inside module and not at import levels anymore, and we don't want to create another S3 client everytime, better keeping the already instanciated one. We could rename `get_s3_client` to `get_or_create_s3_client` for clarity?

udata_hydra/utils/s3.py
```diff
@@ -14,6 +15,27 @@
     "pmtiles": "application/vnd.pmtiles",
 }
 
+_client: "S3Client | None" = None
+
+
+def get_s3_client() -> "S3Client":
```
maudetes: See existing patterns for reusable components in `udata_hydra/context.py`.
I think we could align with it?

udata_hydra/cli_entry.py
maudetes: I wouldn't keep this anymore. It adds complexity and we don't need this due to your refactoring.

maudetes (approved):
Thank you for your refactoring to prevent setting up a S3Client if not needed.

I have set the logger level for `botocore` and `boto3` to INFO in https://github.com/datagouv/hydra/pull/436/changes/a3abbdc0cf3a0a1a1e08248f887c0223b480c9fa, similarly to other libs.
Outside from quiet consideration, we don't want DEBUG logs on these libs.

udata_hydra/utils/s3.py
bolinocroustibat: Good idea, done [here](https://github.com/datagouv/hydra/pull/436/changes/7a68cb625ec090e79b38f7d38dfa6dfe6300c5cb)

maudetes (approved):

PR #438, opened 2026-06-09
refactor(status): migrate catalog status to JSONB per-job model

bolinocroustibat:
### Problem
The current issue: [catalog.status](https://www.google.com/search?q=udata_hydra/migrations/main/20230515_rev8_up_rev9.sql) is a single VARCHAR shared by all pipelines. Parallel jobs overwrite each other (e.g., analyse_csv resets status=None while export_parquet is running).

### New schema for [catalog.status](https://github.com/datagouv/hydra/pull/438/changes#diff-a1f75d8bd450f667934d678612998bd5e1fe13a794a4b55f6d5bbe0aad3931dd)

Each job carries its status and timestamp within the same object, updated in a single atomic query, ensuring there is no risk of desynchronization between two parallel columns.
We also take this opportunity to use a single JSONB column with statuses per job as keys, which replaces `status` AND `[status_since](https://www.google.com/search?q=udata_hydra/migrations/main/20250611_add_status_since_catalog.sql)`.

Nested structure, one key per job domain:
```json
{
  "crawler": {
    "state": "CRAWLING_URL",
    "since": "2026-06-09T12:00:00+00:00"
  },
  "csv": {
    "state": "ANALYSING_CSV",
    "since": "2026-06-09T12:00:00+00:00"
  },
  "parquet": {
    "state": "CONVERTING_TO_PARQUET",
    "since": "2026-06-09T12:01:00+00:00"
  }
}
```

- `state`: existing vocabulary (`ANALYSING_CSV`, `BACKOFF`, etc.)
- `since`: UTC timestamptz, for per-job stuck cleanup
- Idle: key absent from the JSON (no `null` anymore), removal via the JSONB "-" operator
- Idle resource: `{}`:  all the job keys have been removed. We don't need a global "switch to idle" method anymore.

### Indexes
- New `catalog_crawlable_idx` index to replace `status_deleted_idx` to help select batches eligible for crawling.
- New `catalog_status_gin_idx`GIN index for API stats (`get_resources_status_counts` for `status != '{}'`).

### JSONB codec everywhere
We also take this opportunity to use the JSONB codec when initializing the DB connection, so there is no longer a need to use `json.loads` and `json.dumps` to convert from/to strings. Python dicts would be written natively to the DB. For reading, if the data is still in string format, there is a small defensive method that handles both.


udata_hydra/db/codec.py
```diff
@@ -0,0 +1,25 @@
...
+
+def parse_json_value(value: Any, default: Any = None) -> Any:
+    """Parse a JSON/JSONB column value that may already be decoded by asyncpg."""
+    if value is None:
+        return default
+    if isinstance(value, (dict, list)):
+        return value
+    if isinstance(value, str):
+        return json.loads(value)
+    return value
```
ThibaudDauce, line 25: With the codec, jsonb always comes back as a dict, no? The `str` branch can only trigger on a double encoded value, and `update_table_record` still calls `convert_dict_values_to_json` for `Check.update`. Remove it there instead?

udata_hydra/db/resource.py
```diff
@@ -154,24 +232,25 @@ async def delete(
                 q = f"""UPDATE catalog SET deleted = TRUE WHERE resource_id = '{resource_id}';"""
             await connection.execute(q)
 
-    @staticmethod
-    def get_excluded_clause() -> str:
+    @classmethod
+    def get_excluded_clause(cls) -> str:
```
ThibaudDauce, line 236: Why this change? `cls` is not used?

udata_hydra/data_formats/parquet/to_db.py
```diff
@@ -77,8 +77,7 @@ async def parquet_to_db(
         )
 
     if file.resource_id:
-        # Update resource status to INSERTING_IN_DB
-        await Resource.update(file.resource_id, {"status": "INSERTING_IN_DB"})
+        await Resource.set_job_status(file.resource_id, "csv", "INSERTING_IN_DB")
```
ThibaudDauce, line 80:
```suggestion
        await Resource.set_job_status(file.resource_id, "parquet", "INSERTING_IN_DB")
```

Need to add `INSERTING_IN_DB` inside `cls.JOB_STATUSES` too, no?

udata_hydra/data_formats/geojson/__init__.py
```diff
@@ -32,21 +33,27 @@ async def analyse(self, check: dict) -> None:
...
+                data={
+                    "parsing_finished_at": datetime.now(timezone.utc),
+                },
+            )
+        except ParseException as e:
+            check = await handle_parse_exception(e, None, check)  # type: ignore[assignment]
+        finally:
+            await helpers.notify_udata(resource, check)
+            timer.stop()
+            self.path.unlink()
```
ThibaudDauce, line 55:
```suggestion
            self.path.unlink(missing_ok=True)
```

I think the file could already be deleted from the `if delete_input and config.REMOVE_GENERATED_FILES and getattr(data_object, "file_name", None)`, no?

udata_hydra/data_formats/geojson/__init__.py
```diff
@@ -32,21 +33,27 @@ async def analyse(self, check: dict) -> None:
...
+            check = await Check.update(  # type: ignore[assignment]
+                check_id=check["id"],
+                data={
+                    "parsing_finished_at": datetime.now(timezone.utc),
+                },
+            )
+        except ParseException as e:
+            check = await handle_parse_exception(e, None, check)  # type: ignore[assignment]
+        finally:
+            await helpers.notify_udata(resource, check)
```
ThibaudDauce, line 53: Already done in `_run_export_job` no?

udata_hydra/data_formats/geojson/__init__.py
```diff
@@ -32,21 +33,27 @@ async def analyse(self, check: dict) -> None:
...
+            await export_pmtiles(geojson_file=self, check=check)
+            timer.mark("geojson-to-pmtiles")
+            check = await Check.update(  # type: ignore[assignment]
+                check_id=check["id"],
+                data={
+                    "parsing_finished_at": datetime.now(timezone.utc),
+                },
+            )
+        except ParseException as e:
+            check = await handle_parse_exception(e, None, check)  # type: ignore[assignment]
```
ThibaudDauce, line 51: I think `ParseException` is already managed in `_run_export_job`, no?

udata_hydra/schemas/resource.py
```diff
@@ -20,9 +20,13 @@ class ResourceDocumentSchema(Schema):
     harvest = fields.Dict()
 
 
+class JobStatusSchema(Schema):
+    state = fields.Str(required=True)
+    since = fields.DateTime(required=True)
```
ThibaudDauce, line 25: Embed inside JSONB cannot be datetime because JSON doesn't have a datetime type (it's converted to string), this line is crashing the API, you could add this test to reproduce:

```python
async def test_get_resource_with_active_job(setup_catalog, client):
      await Resource.set_job_status(RESOURCE_ID, "crawler", "CRAWLING_URL")

      resp = await client.get(f"/api/resources/{RESOURCE_ID}")
      assert resp.status == 200
      data: dict = await resp.json()
      assert data["status"]["crawler"]["state"] == "CRAWLING_URL"
      assert datetime.fromisoformat(data["status"]["crawler"]["since"])
```

tests/test_analysis/test_analysis_parquet.py
ThibaudDauce, line 88: You should check here for the resource["status"], it fails for the error with the `"csv"` instead of `"parquet"`

```suggestion
    # Check resource status after analysis
    resource = await Resource.get(RESOURCE_ID)
    assert resource is not None
    assert resource["status"] == {}

    # checking check result
```

ThibaudDauce (changes requested):

udata_hydra/data_formats/parquet/to_db.py
ThibaudDauce: You can add `assert resource["status"] == {}` in the parquet test to reproduce the bug (see below)

udata_hydra/db/resource.py
bolinocroustibat: I think this is an old merge issue. Reverted here https://github.com/datagouv/hydra/pull/438/changes/4c907baf7024ca4cf33f7790275db103b9ba4cc5

udata_hydra/data_formats/parquet/to_db.py
bolinocroustibat: > Need to add INSERTING_IN_DB inside cls.JOB_STATUSES too, no?

Already the case here: https://github.com/datagouv/hydra/pull/438/changes#diff-b48977da7dd51be0ffb9dceb039390e3b9dbed2ec763f15a355929d3b9cd27f6R21

ThibaudDauce: Not for `csv` but for `parquet`?

udata_hydra/data_formats/geojson/__init__.py
bolinocroustibat: Generally a good practice to have `missing_ok=True` anyway

udata_hydra/data_formats/parquet/to_db.py
bolinocroustibat: Oh indeed sorry: https://github.com/datagouv/hydra/pull/438/changes/1ab350f16791e2337db6747ef23472f721aac126

tests/test_analysis/test_analysis_parquet.py
bolinocroustibat: Fixed with https://github.com/datagouv/hydra/pull/438/changes/c5c5a2965d51538a42618353b3bffecaef422943, https://github.com/datagouv/hydra/pull/438/changes/07b636137d89eb862e5ddd2fcb419a241b91deba and https://github.com/datagouv/hydra/pull/438/changes/1ab350f16791e2337db6747ef23472f721aac126

udata_hydra/schemas/resource.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/438/changes/06da74ed614339e03a9449e4b23784121d0112de

udata_hydra/data_formats/geojson/__init__.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/438/changes/03ddabc1d98b185a62a4813a3949f52c308b8870

udata_hydra/data_formats/geojson/__init__.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/438/changes/03ddabc1d98b185a62a4813a3949f52c308b8870

udata_hydra/db/codec.py
bolinocroustibat: It's a defensive method in case we still have values that are encoded in JSON, but might be overkill.

bolinocroustibat: Removed indeed

ThibaudDauce:
@bolinocroustibat Maybe a little late, but have you considered creating a new table `resource_job_status(resource_id, job, state, since)` instead of using JSON for this feature? It may be much simpler in requests and atomicity?

udata_hydra/migrations/main/20260609_status_jsonb_catalog.sql
```diff
@@ -0,0 +1,30 @@
...
+ALTER TABLE catalog
+    ALTER COLUMN status SET DEFAULT '{}'::jsonb,
+    ALTER COLUMN status SET NOT NULL;
+
+ALTER TABLE catalog
+    DROP COLUMN status_since;
+
+DROP INDEX IF EXISTS status_deleted_idx;
+
+CREATE INDEX IF NOT EXISTS catalog_status_gin_idx ON catalog USING GIN (status);
```
ThibaudDauce: I think this index is useless

| Requête                   | Index en place                      | Plan                                           | Temps       | Buffers |
| ------------------------- | ----------------------------------- | ---------------------------------------------- | ----------- | ------- |
| stats (PR)                | **GIN** + crawlable                 | `Parallel Seq Scan` + `Filter: status <> '{}'` | 16,7 ms     | 3654    |
| stats (PR)                | **index partiel**                   | `Bitmap Index Scan`                            | **0,34 ms** | **105** |
| stats **sans** le `WHERE` | index partiel                       | `Parallel Seq Scan`                            | 27 ms       | 3654    |
| stats                     | btree `(status, deleted)` sur jsonb | `Parallel Seq Scan`                            | 29 ms       | 3654    |
| comptage idle             | tous                                | `Parallel Seq Scan`                            | 19,6 ms     | 3638    |
| sélection de batch        | crawlable                           | `Bitmap Index Scan on crawlable_idx`           | 8,6 ms      | 3717    |

It's better to have a partial index as follow:

```suggestion
CREATE INDEX IF NOT EXISTS catalog_active_status_idx ON catalog (resource_id) WHERE status <> '{}'::jsonb;
```

This partial index is perfect for the condition `WHERE status != '{}'::jsonb` where most of the table is idle, it allows to filter out all the idle resources. And also add `catalog.status <> '{}'::jsonb` after `f"NOT {CRAWLABLE_STATUS_CLAUSE}"` to use it there too.

But as I said in my previous comment, that's the kind of stuff that can be much simpler with a regular table instead of json…

udata_hydra/db/resource.py
```diff
@@ -180,13 +260,16 @@ async def clean_up_statuses() -> int:
...
-                )
+                SET status = status - stuck.job
+                FROM (
+                    SELECT ca.resource_id, t.key AS job
+                    FROM catalog ca
+                    JOIN checks c ON c.id = ca.last_check
+                    CROSS JOIN LATERAL jsonb_each(ca.status) AS t(key, entry)
+                    WHERE c.created_at < $1
+                      AND (entry->>'since')::timestamptz < $1
+                ) stuck
```
ThibaudDauce: This query only clear one stuck job per call instead of clearing all stuck statuses inside the JSON, something like:


```suggestion
                SET status = status - stuck.jobs
                FROM (
                    SELECT ca.resource_id, array_agg(t.key) AS jobs
                    FROM catalog ca
                    JOIN checks c ON c.id = ca.last_check
                    CROSS JOIN LATERAL jsonb_each(ca.status) AS t(key, entry)
                    WHERE c.created_at < $1 AND (entry->>'since')::timestamptz < $1
                    GROUP BY ca.resource_id
                ) stuck
```

(same thing, could be simpler without JSON)

udata_hydra/crawl/check_resources.py
```diff
@@ -93,9 +93,8 @@ async def check_resource(
     if should_backoff:
         log.info(f"backoff {domain} ({reason})")
         # skip this URL, it will come back in a next batch
-        await Resource.update(
-            str(resource["resource_id"]), data={"status": "BACKOFF", "priority": False}
-        )
+        await Resource.set_job_status(str(resource["resource_id"]), "crawler", "BACKOFF")
+        await Resource.update(str(resource["resource_id"]), data={"priority": False})
```
ThibaudDauce: Maybe `set_job_status` could accept an additional  `data` dict to avoid having two `UPDATE` instead of one (better perfs) and to avoid race condition between the two?

udata_hydra/crawl/preprocess_check_data.py
```diff
@@ -90,7 +89,8 @@ async def preprocess_check_data(dataset_id: str, check_data: dict) -> tuple[dict
     # Update resource following check:
     # Reset resource status so that it's not forbidden to be checked again.
     # Reset priority so that it's not prioritised anymore.
-    await Resource.update(check_data["resource_id"], data={"status": None, "priority": False})
+    await Resource.clear_job_status(check_data["resource_id"], "crawler")
+    await Resource.update(check_data["resource_id"], data={"priority": False})
```
ThibaudDauce: Same here, if `clear_job_status` accepts a additional dict for `priority` we avoid two UPDATE and a potential race condition between the two queries.

udata_hydra/data_formats/parquet/__init__.py
```diff
@@ -69,45 +70,58 @@ async def analyse(self, check: dict, debug_insert: bool = False) -> "Table|None"
...
+                },
+            )
+            return table
+        except ParseException as e:
+            check = await handle_parse_exception(
+                e, table.table_name if table is not None else None, check
+            )  # type: ignore[assignment]
+            return None
+        finally:
+            await helpers.notify_udata(resource, check)
```
ThibaudDauce: What happen if the `notify_udata` fail? The `clear_job_status` is never called? Maybe we should call `clear_job_status` before, or improve the robustness here?

udata_hydra/data_formats/ogc/__init__.py
```diff
@@ -245,7 +243,7 @@ async def analyse(cls, check: dict) -> OgcMetadata | None:
             if resource and check_id:
                 await helpers.notify_udata(resource, check)
```
ThibaudDauce: Same here. What happen if the `notify_udata` fail? The `clear_job_status` is never called? Maybe we should call `clear_job_status` before, or improve the robustness here?

udata_hydra/analysis/exports.py
```diff
@@ -72,7 +73,8 @@ async def _run_export_job(
             resource = await Resource.get(data_object.resource_id)
             if resource is not None and check_out is not None:
                 await helpers.notify_udata(resource, check_out)
```
ThibaudDauce: Same here. What happen if the `notify_udata` fail? The `clear_job_status` is never called? Maybe we should call `clear_job_status` before, or improve the robustness here?

udata_hydra/data_formats/parquet/__init__.py
```diff
@@ -69,45 +70,58 @@ async def analyse(self, check: dict, debug_insert: bool = False) -> "Table|None"
...
+            check = await Check.update(  # type: ignore[assignment]
+                check_id=check["id"],
+                data={
+                    "parsing_finished_at": datetime.now(timezone.utc),
+                },
+            )
+            return table
+        except ParseException as e:
+            check = await handle_parse_exception(
+                e, table.table_name if table is not None else None, check
```
ThibaudDauce: `table` is always `None` here because `ParseException` is raised by `self.to_db` before the `table` assignment. So the table is never cleanup if `copy_records_to_table` fail (after the create table)?

Is this better? Or use `e.table_name` inside `handle_parse_exception`? But `_run_export_job` doesn't set it on the exception, it would need a fix too.
```suggestion
                e, e.table_name, check
```

tests/test_crawl/test_crawl.py
```diff
@@ -695,24 +690,38 @@ async def test_content_type_from_header(content_type):
...
+    else:
+        await Resource.set_job_status(
+            resource_id, Resource.job_for_state(resource_status), resource_status
+        )
+
+
+@pytest.mark.parametrize(
+    "resource_status",
+    [s for states in Resource.JOB_STATUSES.values() for s in states],
+)
```
ThibaudDauce: We test twice the `csv / INSERTING_IN_DB` but never the `parquet / INSERTING_IN_DB` 

```suggestion
@pytest.mark.parametrize(
      "job,state",
      [(job, s) for job, states in Resource.JOB_STATUSES.items() for s in sorted(states)],
  )
```

`job_for_state` cannot work because `state` is not unique. It can be removed and tests should be fixed.

ThibaudDauce (changes requested):
Still thinking after my review that some stuff should be simpler with a table instead of JSON if it's not too much work to switch the model…

bolinocroustibat:
> @bolinocroustibat Maybe a little late, but have you considered creating a new table `resource_job_status(resource_id, job, state, since)` instead of using JSON for this feature? It may be much simpler in requests and atomicity?

Yes, that would improve readability, and maybe performance as we wouldn't have to update rows in a big table.
Here is a first PR: https://github.com/datagouv/hydra/pull/474

bolinocroustibat:
Replaced by https://github.com/datagouv/hydra/pull/474

PR #441, opened 2026-06-15
fix(cli): make fd var bound (initialize to None)

maudetes:
Currently, our crons fail with
```
UnboundLocalError: cannot access local variable 'fd' where it is not associated 
with a value
```

The underlying error may be related to the temporary file creation.


bolinocroustibat (approved):
LGTM, thanks for quick fix!

PR #443, opened 2026-06-15
Un-revert data formats and fix file routing

Pierlou:
Guess who's back https://github.com/datagouv/hydra/pull/430
Includes a fix for the files' routing: `DataFormat` instances now only have the file's name as an attribute, and the function `storage_path` allows to locate it at run time (done in [this commit](https://github.com/datagouv/hydra/pull/443/changes/ef2bbcaa594740f4d0d57c5c26f473a2d7489783)). All converted files are also stored using `storage_path`, so they end up in the same (temporary) directory as the downloaded files


udata_hydra/utils/file.py
```diff
@@ -126,3 +126,9 @@ def remove_remainders(resource_id: str, extensions: list[str]) -> None:
     for ext in extensions:
         if os.path.exists(f"{resource_id}.{ext}"):
             os.remove(f"{resource_id}.{ext}")
+
+
+def true_path(file_name: str):
+    if file_name.startswith("tests/data/"):
+        return file_name
+    return f"{config.TEMPORARY_DOWNLOAD_FOLDER or '/tmp'}/{file_name}"
```
bolinocroustibat: I would use `tempfile.gettempdir()` as the fallback instead of `/tmp`, so that it adapts to the OS/environment. Several OS like Windows and MacOS don't use `/tmp`, more robust. Let me know what you think

udata_hydra/utils/file.py
```diff
@@ -126,3 +126,9 @@ def remove_remainders(resource_id: str, extensions: list[str]) -> None:
     for ext in extensions:
         if os.path.exists(f"{resource_id}.{ext}"):
             os.remove(f"{resource_id}.{ext}")
+
+
+def true_path(file_name: str):
```
bolinocroustibat: The name `true_path` is not super explicit (plus it returns a `str` not a `Path` which is not clear from the name).
What about `temporary_file_path` if it returns a `Path`  - or `temporary_file_name` if it returns a `str`?

udata_hydra/utils/file.py
bolinocroustibat: Could it return a `Path` type?
I really liked the usage of `Path` types instead of `str` in the previous commits, makes everything safer (and more pythonic).

bolinocroustibat: No need to create the directory in case it doesn't exist?

Something like:

```python
    if config.TEMPORARY_DOWNLOAD_FOLDER:
        folder = Path(config.TEMPORARY_DOWNLOAD_FOLDER)
    else:
        folder = Path(tempfile.gettempdir()) # or `folder = Path("/tmp")` if we want to use `/tmp`
    folder.mkdir(parents=True, exist_ok=True)
    return folder/file_name  # or `return f"{folder.name}/{file_name}"` if we can't return a `Path`
```

bolinocroustibat (changes requested):
Thank you for this! I have several remarks:

- We should align [`_load_catalog` in `catalog.py`](https://github.com/datagouv/hydra/blob/c4d1af7746d5339660252c28b7bd77d8e76bc225/udata_hydra/cli/crawl.py#L55) and and [`_download_resource_cli` in `crawl.py`](https://github.com/datagouv/hydra/blob/c4d1af7746d5339660252c28b7bd77d8e76bc225/udata_hydra/cli/catalog.py#L48) to also use `true_path` instead of `config.TEMPORARY_DOWNLOAD_FOLDER` to finally have full consistency about the temp files paths all over the codebase

- I'm not sure why we can't use `Path` objects anymore in `DataFormat` as property for the file? I might have misunderstood something. Is it because `DataFormat.file_name` is used in all the conversions lifecycle?
If not, we could initialize it using `true_path(...)` if necessary.  Unless impossible, using this type for the file property all over the codebase really improves safety (and more pythonic), we should use it as much as we can.
If we can't, could we still resolve `true_path()` to return a `Path` anyway?

- Some other remarks/questions

udata_hydra/utils/file.py
bolinocroustibat: NIT/very minor: I would put that method on top of `file.py` since it's the most called method of `file.py` from the codebase, easier to reach when reading the codebase

udata_hydra/data_formats/data_format.py
```diff
@@ -0,0 +1,61 @@
+import json
+import os
+from abc import ABC, abstractmethod
+
+from udata_hydra.utils import true_path
+
+
+class DataFormat(ABC):
+    file_name: str
```
bolinocroustibat, line 10: Too bad we cannot keep a `Path` type arg here after resolving it with `true_path`?

```python
class DataFormat(ABC):
    path: Path
    def __init__(
        self,
        *,
        file_name: str | None = None,
        table_name: str | None = None,
        inspection: dict | None = None,
        resource_id: str | None = None,
        dataset_id: str | None = None,
    ) -> None:
        if file_name:
            self.path = true_path(file_name)
```


If not possible, maybe we can have a helpful property?
```python
class DataFormat(ABC):
    file_name: str
    @property
    def path(self) -> Path:
        return true_path(self.file_name)  # true_path returns Path
```
(not sure that second option simplifies, just an idea)

udata_hydra/utils/file.py
Pierlou: Agreed, changed in https://github.com/datagouv/hydra/pull/443/commits/902660297427515bf7e5019004b7fe19b327d538

Pierlou: I believe there would never be a case of the parent folder not existing, it's either specified in the config (and in this case I'd say it's on the infra's side that this should be managed) or the tmp folder of the OS, which always exists?

udata_hydra/data_formats/data_format.py
Pierlou: That's actually a better alternative I think, I'll try to go for that

udata_hydra/utils/file.py
Pierlou: It's a bit ambiguous I agree, though the routing for `tests/data/` files would be misleading if mentionning `temporary` in the function's name 🤔

udata_hydra/data_formats/data_format.py
Pierlou: Done in https://github.com/datagouv/hydra/pull/443/commits/1cb38948d59cc407d47da883a8a34ac0933bca21, that indeed looks way better, thanks for the suggestion!

tests/conftest.py
```diff
@@ -110,7 +111,7 @@ def setup():
         WEBHOOK_ENABLED=True,
         SENTRY_DSN=None,
         # Align download_resource with read_or_download_file basename lookup (or "" uses OS temp vs /tmp).
-        TEMPORARY_DOWNLOAD_FOLDER="/tmp",
+        TEMPORARY_DOWNLOAD_FOLDER=true_path("").as_posix(),
```
bolinocroustibat: No need to `.as_posix` if you want, `tempfile` accepts a `Path` directly since Python 3.6+:
https://docs.python.org/3/library/tempfile.html -> _Changed in version 3.6: The dir parameter now accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object)_

Just a style issue

udata_hydra/cli/catalog.py
```diff
@@ -44,9 +44,7 @@ def iter_with_progressbar_or_quiet(rows, quiet):
 
         try:
             log.info(f"Downloading resources catalog from {url}...")
-            with NamedTemporaryFile(
-                dir=config.TEMPORARY_DOWNLOAD_FOLDER or None, delete=False
-            ) as fd:
+            with NamedTemporaryFile(dir=true_path("").as_posix(), delete=False) as fd:
```
bolinocroustibat: No need to `.as_posix` if you want, `tempfile` accepts a `Path` directly since Python 3.6+:
https://docs.python.org/3/library/tempfile.html -> _Changed in version 3.6: The dir parameter now accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object)_

Just a style issue

udata_hydra/utils/file.py
```diff
@@ -37,7 +37,7 @@ def compute_checksum_from_file(filename: str) -> str:
 def extract_gzip(file_path: str) -> IO[bytes]:
     with gzip.open(file_path, "rb") as gz_file:
         with tempfile.NamedTemporaryFile(
-            dir=config.TEMPORARY_DOWNLOAD_FOLDER or None, mode="wb", delete=False
+            dir=true_path("").as_posix(), mode="wb", delete=False
```
bolinocroustibat: No need to `.as_posix` if you want, `tempfile` accepts a `Path` directly since Python 3.6+:
https://docs.python.org/3/library/tempfile.html -> _Changed in version 3.6: The dir parameter now accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object)_

Just a style issue

udata_hydra/utils/file.py
```diff
@@ -60,9 +60,7 @@ async def download_resource(
     ):
         raise IOException("File too large to download", url=url)
 
-    tmp_file = tempfile.NamedTemporaryFile(
-        dir=config.TEMPORARY_DOWNLOAD_FOLDER or None, delete=False
-    )
+    tmp_file = tempfile.NamedTemporaryFile(dir=true_path("").as_posix(), delete=False)
```
bolinocroustibat: No need to `.as_posix` if you want, `tempfile` accepts a `Path` directly since Python 3.6+:
https://docs.python.org/3/library/tempfile.html -> _Changed in version 3.6: The dir parameter now accepts a [path-like object](https://docs.python.org/3/glossary.html#term-path-like-object)_

Just a style issue

udata_hydra/cli/crawl.py
```diff
@@ -52,8 +52,7 @@ async def _download_resource_cli(resource_id: str, output_dir: str | None = None
     try:
         tmp_file, file_extension = await download_resource(resource["url"])
         output_path = (
-            Path(output_dir or config.TEMPORARY_DOWNLOAD_FOLDER or ".")
-            / f"{resource_id}{file_extension}"
+            Path(output_dir or true_path("").as_posix()) / f"{resource_id}{file_extension}"
```
bolinocroustibat: Same here, I think it accepts Path objects

udata_hydra/utils/file.py
bolinocroustibat: NIT: Still not 100% satisfied with the name, here are some other ideas after yesterday's suggestions:
- `work_path`?
- `storage_path`? (my favorite)
- `resolve_path`?
- `temp_path`? (very clear but I understand it's not only for temp files?)

bolinocroustibat (commented):
LGTM, thank you! Some non-blocking suggestions

tests/conftest.py
Pierlou: As it's patching the config I was making sure to push an actual string, to prevent any unexpected behaviour

udata_hydra/utils/file.py
Pierlou: Let's go for `storage_path`, I like it too 🤝

PR #446, opened 2026-06-16
Ensure we pass dicts to enqueue

Pierlou:
To prevent [this error](https://errors.data.gouv.fr/organizations/sentry/issues/294986)


bolinocroustibat (approved):
Ah yes there is always an inconsistency in the codebase for the type of `Check`, between `Record` and `dict`.

For info I've been [trying to fix while trying to stay "elegant" (by keeping the CRUD methods in `db/check.py` to keep returning `Record`](https://github.com/datagouv/hydra/pull/331)), but it's not that easy, so I gave up for now.

Anyway, well spotted, LGTM

PR #447, opened 2026-06-16
Don't try to delete a file from a table

Pierlou:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/295216

### DIFF
--- a/udata_hydra/analysis/exports.py
+++ b/udata_hydra/analysis/exports.py
@@ -43,7 +43,11 @@ async def _run_export_job(
                     f"{df_name.lower()}_size": output.filesize,
                 },
             )
-            if delete_input and config.REMOVE_GENERATED_FILES and getattr(data_object, "file_name"):
+            if (
+                delete_input
+                and config.REMOVE_GENERATED_FILES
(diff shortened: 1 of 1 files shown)


udata_hydra/analysis/exports.py
```diff
@@ -43,7 +43,7 @@ async def _run_export_job(
                     f"{df_name.lower()}_size": output.filesize,
                 },
             )
-            if delete_input and config.REMOVE_GENERATED_FILES:
+            if delete_input and config.REMOVE_GENERATED_FILES and getattr(data_object, "file_name"):
                 data_object.path.unlink()
```
bolinocroustibat, line 47: If file_name is malformed or empty it will be OK?

bolinocroustibat (approved):
LGTM!

udata_hydra/analysis/exports.py
Pierlou: Indeed not, I added a default in https://github.com/datagouv/hydra/pull/447/commits/358845e722d3baffd815c2edb99d6d4057cd3ffb

PR #448, opened 2026-06-16
Use the correct kwarg in Geojson + PMTiles export

Pierlou:
Fix https://errors.data.gouv.fr/organizations/sentry/issues/295214


bolinocroustibat (approved):
LGTM!

PR #449, opened 2026-06-16
test: add file.py unit tests

bolinocroustibat:
Add unit tests for `file.py`.

Test coverage is up from 83,8% to 85.1%

Later on, once `remove_remainders` is adjusted, we can also maybe add an unit test for `remove_remainders` in this PR.


Pierlou (approved):
Thanks 🚀

PR #450, opened 2026-06-16
Remove remainders targets `storage_path`

Pierlou:


udata_hydra/utils/file.py
```diff
@@ -129,5 +129,6 @@ async def download_file(url: str, fd):
 def remove_remainders(resource_id: str, extensions: list[str]) -> None:
     """Delete potential remainders from process that crashed"""
     for ext in extensions:
-        if os.path.exists(f"{resource_id}.{ext}"):
-            os.remove(f"{resource_id}.{ext}")
+        path = storage_path(f"{resource_id}.{ext}")
+        if path.exists():
```
bolinocroustibat: I think we can use `path.unlink(missing_ok=True)` to pass if it doens't exist, instead of checking `if path.exists()`

```python
storage_path(f"{resource_id}.{ext}").unlink(missing_ok=True)
```

bolinocroustibat (approved):
LGTM! Just a code style remark

tests/test_utils/test_file.py
bolinocroustibat, line 33: Actually I just realized that if we want to use `Path` everywhere and avoid another import (`import os`), we could do:

```suggestion
    extracted = extract_gzip(str(gz_path))
    extracted_path = Path(extracted.name)

    try:
        assert extracted_path.parent == storage_path("")
        with open(extracted_path, "rb") as f:
            assert f.read() == b"col\nval"
    finally:
        extracted_path.unlink()
```

bolinocroustibat (commented):
LGTM, just two remarks, one not related to your changes

udata_hydra/utils/file.py
Pierlou: Done in https://github.com/datagouv/hydra/pull/450/commits/c711b6d41f232db38060a5d5e47658ecf1424221 ✨

tests/test_utils/test_file.py
Pierlou: Sure, maybe we do that in a following PR?

bolinocroustibat: Yes!

PR #455, opened 2026-06-24
Add custom NA values for tabular data

Pierlou:


maudetes (commented):
👀 I haven't followed the discussion on csv-detective, but we're going for a list of known NA values to convert to null?

udata_hydra/utils/casting.py
```diff
@@ -4,14 +4,14 @@
...
-from udata_hydra.utils.reader import Reader
+from udata_hydra.utils import NA_VALUES, Reader
 
 log = logging.getLogger("udata-hydra")
 
 
 def _smart_cast(_type: str, value, cast_json: bool = True, failsafe: bool = False) -> Any:
     try:
-        if value is None or value == "":
+        if value is None or value == "" or value in NA_VALUES:
```
ThibaudDauce, line 14: We cannot do that for all types because `NA` and `NC` are valid country codes. We should do that only for non-string columns I think

Pierlou: I believe we can't have it all here: in `csv-detective` (as in `pandas`) there is no subtility in the loading of NA values, so **all** values that are in the NA list will be loaded as `NaN` and therefore not tested for any format (which is not blocking as hydra keeps `skipna=True`)

PR #456, opened 2026-06-24
feat(db): add transaction for csv-to-db

maudetes:
The objectives are :
- to prevent a table deletion if the following creation or upserting fails
- to prevent users from getting partially fed table during `copy_records_to_table` (that can take up to some minutes)

~~We could have gone with a table swap by changing names, but since we have indexes, it would require updating the indexes name and sounded more cumbersome. Let me know if there are other considerations.~~
We went for a table swap to prevent READ lock during the table creation and insertion.

### DIFF
--- a/tests/test_analysis/test_analysis_parquet.py
+++ b/tests/test_analysis/test_analysis_parquet.py
@@ -9,7 +9,9 @@ from unittest.mock import patch
 import pandas as pd
 import pyarrow.parquet as pq
 import pytest
+from slugify import slugify
 
+from udata_hydra import config
 from udata_hydra.analysis import helpers
 from udata_hydra.data_formats import Parquet
 from udata_hydra.utils import ParseException, storage_path
@@ -212,3 +214,48 @@ async def test_parquet_to_db_create_table_failure_raises_parse_exception(
     assert res[0]["val"] == "original-data"
 
     await db.execute(f'DROP TABLE IF EXISTS "{table_name}"')
+
+
+async def test_parquet_to_db_indexes_renamed(db, clean_db, fake_check):
+    """Indexes created with shadow table names are renamed to the real table name."""
+    check = await fake_check()
+    table_name = hashlib.md5(check["url"].encode("utf-8")).hexdigest()
+
+    df = pd.DataFrame(
+        {"name": ["Alice", "Bob"], "score": [100, 90], "category": ["Science", "Arts"]}
+    )
+    inspection = {
+        "columns": {
+            "name": {"python_type": "string", "format": set()},
+            "score": {"python_type": "int", "format": set()},
+            "category": {"python_type": "string", "format": set()},
+        },
+        "total_lines": 2,
+    }
+    table_indexes = {"name": "index", "category": "index"}
+    with NamedTemporaryFile(suffix=".parquet", dir=storage_path("")) as fp:
+        df.to_parquet(fp.name)
+        file = Parquet(file_name=os.path.basename(fp.name), inspection=inspection)
+        table = await file.to_db(check=check, table_indexes=table_indexes)
+
+    assert table.table_name == table_name
+
+    rows = await db.fetch(
+        "SELECT indexname FROM pg_indexes WHERE tablename = $1 AND schemaname = $2",
+        table_name,
+        config.DATABASE_SCHEMA,
+    )
+    index_names = [r["indexname"] for r in rows]
+
+    assert not any(name.startswith(f"{table_name}_s") for name in index_names), (
+        f"Index names still contain shadow suffix: {index_names}"
+    )
+
+    for col in ("name", "category"):
+        expected = f"{table_name}_{slugify(col)}_idx"
+        assert expected in index_names, f"Expected index {expected} not found in {index_names}"
+
+    res = list(await db.fetch(f'SELECT * FROM "{table_name}" ORDER BY name'))
+    assert len(res) == 2
+    assert res[0]["name"] == "Alice"
+    assert res[1]["name"] == "Bob"
--- a/tests/test_analysis/test_analysis_parquet.py
+++ b/tests/test_analysis/test_analysis_parquet.py
@@ -1,6 +1,9 @@
+import hashlib
 import io
 import json
+import os
 from datetime import datetime
+from tempfile import NamedTemporaryFile
 from unittest.mock import patch
 
 import pandas as pd
@@ -9,7 +12,7 @@ import pytest
 
 from udata_hydra.analysis import helpers
 from udata_hydra.data_formats import Parquet
-from udata_hydra.utils import ParseException
+from udata_hydra.utils import ParseException, storage_path
 
 pytestmark = pytest.mark.asyncio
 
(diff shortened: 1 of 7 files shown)


udata_hydra/data_formats/csv_like/to_db.py
```diff
@@ -64,47 +64,38 @@ async def csv_to_db(
...
-    await db.execute(q)
-
-    # Create table
-    q = compute_create_table_query(table_name=table_name, columns=columns, indexes=table_indexes)
+    step = "create_table_query"
     try:
-        await db.execute(q)
+        async with db.acquire() as conn:
+            async with conn.transaction():
+                await conn.execute(f'DROP TABLE IF EXISTS "{table_name}"')
```
bolinocroustibat: Since we actually have two queries between the step "create_table_query" and "copy_records_to_table", could/should we have more granular steps for ParseException?
for example:
- step = "drop_table"
- step = "create_table"
- step = "copy_records_to_table"

Would that be useful?

udata_hydra/data_formats/csv_like/to_db.py
```diff
@@ -64,47 +64,38 @@ async def csv_to_db(
...
-    # this inserts rows from iterator one by one, slow but useful for debugging
-    else:
-        bar = ProgressBar(total=inspection["total_lines"])
-        for r in bar.iter(iter_tabular_rows(file.path, inspection, cast_json=False)):
-            data = {k: v for k, v in zip(inspection["columns"], r)}
-            # NB: possible sql injection here, but should not be used in prod
-            q = compute_insert_query(table_name=table_name, data=data, returning="__id")
-            await db.execute(q, *data.values())
-
     check = await Check.update(check["id"], {"parsing_table": table_name})  # type: ignore[assignment]
```
bolinocroustibat: Not really related to this PR: We could also have some king of atomic transaction or try/expect logic for `Check.update` and `insert_tables_index_entry` to make sure one cannot go with the other one? Just an idea.

tests/test_analysis/test_analysis_parquet.py
```diff
@@ -136,51 +139,76 @@ async def test_parquet_to_db_rejects_too_long_column_name(fake_check):
...
-    with pytest.raises(ParseException) as exc:
-        await file.to_db(check=check)
+    df = pd.DataFrame({"name": ["test"]})
+    with NamedTemporaryFile(suffix=".parquet", dir=storage_path("")) as fp:
+        df.to_parquet(fp.name)
+        file = Parquet(file_name=os.path.basename(fp.name), inspection=inspection)
+        with pytest.raises(ParseException) as exc:
+            await file.to_db(check=check)
+
     assert exc.value.step == "create_table_query"
```
bolinocroustibat, line 210: Very nitpick (not sure it's worth it since it complicates the test code), we could also thet the step in the ParseException with:

```pyhton
with pytest.raises(ParseException) as exc:
    await file.to_db(check=check)
assert exc.value.step == "create_table_query"
```

tests/test_analysis/test_analysis_parquet.py
```diff
@@ -136,51 +139,76 @@ async def test_parquet_to_db_rejects_too_long_column_name(fake_check):
...
-    with pytest.raises(ParseException) as exc:
-        await file.to_db(check=check)
+    df = pd.DataFrame({"name": ["test"]})
+    with NamedTemporaryFile(suffix=".parquet", dir=storage_path("")) as fp:
+        df.to_parquet(fp.name)
+        file = Parquet(file_name=os.path.basename(fp.name), inspection=inspection)
+        with pytest.raises(ParseException) as exc:
+            await file.to_db(check=check)
+
     assert exc.value.step == "copy_records_to_table"
```
bolinocroustibat, line 175: Very nitpick (not sure it's worth it since it complicates the test code), we could also thet the step in the ParseException with:

```pyhton
with pytest.raises(ParseException) as exc:
    await file.to_db(check=check)
assert exc.value.step == "copy_records_to_table"
```

bolinocroustibat (approved):
Great improvement, thank you!
A few comments/questions, not sure if they're worth it.

By the way maybe we could apply this wrapping logic of several DB transaction in a single atomic transaction in some other places in the Hydra code, to improve data consistency in case of errors? I'll have a look on that.

tests/test_analysis/test_analysis_parquet.py
maudetes: I'm unsure of what you're saying, but we're testing `assert exc.value.step == "create_table_query"` on line 208

udata_hydra/data_formats/csv_like/to_db.py
maudetes: Agreed, I think it can be done in a different PR

udata_hydra/data_formats/csv_like/to_db.py
maudetes: done in https://github.com/datagouv/hydra/pull/456/changes/559ad35453e7cd2989d029d96e1480674c1deef6

tests/test_analysis/test_analysis_parquet.py
maudetes: added in https://github.com/datagouv/hydra/pull/456/changes/559ad35453e7cd2989d029d96e1480674c1deef6

tests/test_analysis/test_analysis_parquet.py
maudetes: added in https://github.com/datagouv/hydra/pull/456/changes/559ad35453e7cd2989d029d96e1480674c1deef6

udata_hydra/cli/purge.py
```diff
@@ -167,3 +167,49 @@ def purge_selected_csv_tables(
...
+
+
+async def _purge_shadow_tables(
+    quiet: bool = False,
+) -> None:
+    """Drop all leftover shadow tables"""
+    with quiet_logs(enabled=quiet):
+        conn = await connection(db_name="csv")
+        res: list[Record] = await conn.fetch(
+            f"""SELECT tablename FROM pg_catalog.pg_tables
```
bolinocroustibat, line 179: I'm not sure I understand the regex and how it select the shadow tables and not the others.
I think a comment explaining what is a shadow table and how they are named and how the regex works would help a lot for maintenance!
Not sure if this comment should go here or on `to_db.py` though, or both.

Even better: we could extarct the pattern in a shared constant (`SHADOW_TABLE_SUFFIX_RE = ...`) to avoid any drift, and comment it there maybe. What do you think?

udata_hydra/data_formats/csv_like/to_db.py
```diff
@@ -1,107 +1,134 @@
...
+                    data = {k: v for k, v in zip(inspection["columns"], r)}
+                    q = compute_insert_query(
+                        table_name=shadow_table_name, data=data, returning="__id"
+                    )
+                    await conn.execute(q, *data.values())
+
+        step = "swap_tables"
+        async with db.acquire() as conn:
+            async with conn.transaction():
+                await conn.execute(f'DROP TABLE IF EXISTS "{table_name}"')
```
bolinocroustibat, line 101: Nitpick: we could have more steps names here for even better debug or is it too much?

bolinocroustibat (approved):
LGTM, well done, that's a tricky one!
A few questions remarks in the PR.

Also, should we add a test in `test_purge.py` to test the deletion of the shadow tables? (i.e. create a fake shadow, purge, verify that it disappears and verify real tables stay)

udata_hydra/data_formats/parquet/to_db.py
```diff
@@ -1,138 +1,156 @@
...
+                    data = {k: v for k, v in zip(pqf.schema.names, r)}
+                    q = compute_insert_query(
+                        table_name=shadow_table_name, data=data, returning="__id"
+                    )
+                    await conn.execute(q, *data.values())
+
+        step = "swap_tables"
+        async with db.acquire() as conn:
+            async with conn.transaction():
+                await conn.execute(f'DROP TABLE IF EXISTS "{table_name}"')
```
bolinocroustibat, line 123: Nitpick (same): we could have more steps names here for even better debug or is it too much?

udata_hydra/data_formats/csv_like/to_db.py
```diff
@@ -1,115 +1,134 @@
...
+                await conn.execute(f'ALTER TABLE "{shadow_table_name}" RENAME TO "{table_name}"')
```
bolinocroustibat, line 102: Between those two steps, the table doesn't exist. It should be extremely short time, but in this very little time there could be a 404 response.
Either we keep it in mind and live with it, either we do some kind of atomic transaction here?

EDIT: Sorry actually both are already in the same transaction so it should be fine! The real downtime window is the lock wait, not a missing table. I guess we can live with that! Or plan a special HTTP error code for this case later? Just an idea.

udata_hydra/cli/purge.py
```diff
@@ -167,3 +167,49 @@ def purge_selected_csv_tables(
...
+            WHERE schemaname = '{config.DATABASE_SCHEMA}'
+            AND tablename ~ '^[0-9a-f]{{32}}_s[0-9a-f]{{4}}$'"""
```
bolinocroustibat, line 181: The shadow table naming convention could be centralized somewhere maybe?

tests/test_conversion/test_csv_to_db.py
```diff
@@ -111,11 +115,120 @@ async def test_percentage_column(db, clean_db, fake_check):
...
+            await file.to_db(check=check)
+
+        assert exc.value.step == "copy_records_to_table"
+
+    # Old table and data must survive
+    res = await db.fetch(f'SELECT * FROM "{table_name}"')
+    assert len(res) == 1
+    assert res[0]["val"] == "original-data"
+
+    await db.execute(f'DROP TABLE IF EXISTS "{table_name}"')
```
bolinocroustibat, line 234: tests cover create and copy, but not a failure while we do a "swap_tables". Postgres rollback should protect from any failure, but since it's the critical part, should we add that?

udata_hydra/data_formats/csv_like/to_db.py
```diff
@@ -1,115 +1,134 @@
...
+                        new_name = f"{table_name}{row['indexname'][len(shadow_table_name) :]}"
+                        if new_name != row["indexname"]:
+                            q = f'"{config.DATABASE_SCHEMA}"."{row["indexname"]}"'
+                            await conn.execute(f'ALTER INDEX IF EXISTS {q} RENAME TO "{new_name}"')
+    except Exception as e:
+        try:
+            async with db.acquire() as conn:
+                await conn.execute(f'DROP TABLE IF EXISTS "{shadow_table_name}"')
+        except Exception:
+            log.warning("Failed to clean up shadow table %s", shadow_table_name)
```
bolinocroustibat, line 119: Let's use f-strings interpolation here, for consistency and readability!

udata_hydra/data_formats/parquet/to_db.py
```diff
@@ -1,138 +1,156 @@
...
+                        new_name = f"{table_name}{row['indexname'][len(shadow_table_name) :]}"
+                        if new_name != row["indexname"]:
+                            q = f'"{config.DATABASE_SCHEMA}"."{row["indexname"]}"'
+                            await conn.execute(f'ALTER INDEX IF EXISTS {q} RENAME TO "{new_name}"')
+    except Exception as e:
+        try:
+            async with db.acquire() as conn:
+                await conn.execute(f'DROP TABLE IF EXISTS "{shadow_table_name}"')
+        except Exception:
+            log.warning("Failed to clean up shadow table %s", shadow_table_name)
```
bolinocroustibat, line 141: Let's use f-strings interpolation here, for consistency and readability!

udata_hydra/data_formats/csv_like/to_db.py
maudetes: I think there shouldn't be a 404 indeed but a very short lock wait, but no downtime. Once the transaction is done (which should be almost instant), it is released and the client get an answer.

PR #457, opened 2026-06-30
fix: add dummy migration of resources_exceptions

abulte:


udata_hydra/migrations/csv/20260630_add_resources_exceptions.sql
```diff
@@ -0,0 +1,11 @@
+-- creates the table from `main/20240827_add_resources_exceptions_table.sql`
+-- which migration has been documented in `csv/20250610_migrate_resources_exception.sql` but is a noop
+-- this helps having a working env from scratch but does not apply on prod (IF NOT EXIST)
```
abulte:
```suggestion
-- this helps having a working env from scratch but does not apply on prod (IF NOT EXISTS)
```

maudetes (approved):
Thank you for this fix!

PR #460, opened 2026-07-06
fix: check head request fallback

bolinocroustibat:
When crawling resource `ac59a0f5-fa83-4b82-bf12-3c5806d4f19f` (`monactiviteformation.emploi.gouv.fr`), Hydra’s HEAD request can succeed with HTTP 200 while the response is actually a short WAF page (text/html, e.g. “Request Rejected”, in our case text/html, content-length: 247) instead of the CSV file. `has_nice_head()` treated that as a valid HEAD response, so Hydra skipped the existing GET fallback and recorded a misleading successful check.

This change tightens `has_nice_head()`: a HEAD response with text/html and a small Content-Length (< 4 KB) is now considered unreliable, triggering the existing `HEAD`->`GET` fallback in check_resource. A regression test covers the case.

### DIFF
--- a/udata_hydra/crawl/helpers.py
+++ b/udata_hydra/crawl/helpers.py
@@ -50,6 +50,13 @@ def has_nice_head(resp) -> bool:
         return False
     if not any([k in resp.headers for k in ("content-length", "last-modified")]):
         return False
+    content_type = resp.headers.get("content-type", "").lower()
+    if content_type.startswith("text/html"):
+        try:
+            if int(resp.headers.get("content-length", 0)) < 4096:
+                return False
+        except (TypeError, ValueError):
+            return False
     return True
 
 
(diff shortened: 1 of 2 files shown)


udata_hydra/crawl/helpers.py
```diff
@@ -50,6 +50,13 @@ def has_nice_head(resp) -> bool:
         return False
     if not any([k in resp.headers for k in ("content-length", "last-modified")]):
         return False
+    content_type = resp.headers.get("content-type", "").lower()
+    if content_type.startswith("text/html"):
+        try:
+            if int(resp.headers.get("content-length", 0)) < 4096:
```
ThibaudDauce: Maybe put `4096` inside a config? And maybe add a comment explaining why we do that?

ThibaudDauce (approved):

udata_hydra/crawl/helpers.py
bolinocroustibat: Why not, done here: https://github.com/datagouv/hydra/pull/460/changes/256a13cc0e013b1e07928736d0993a5c4ec81037

PR #465, opened 2026-07-17
refactor: replace Marshmallow with Pydantic

### LINKED ISSUE #150: [hydra] Use Pydantic instead of Marshmallow

bolinocroustibat:
Close #150 

Replace the data validation library [Marshmallow](https://marshmallow.readthedocs.io/en/stable/) with [Pydantic](https://github.com/pydantic/pydantic), a more modern and widely adopted alternative in the Python ecosystem.

Why:
- More lively ecossytem
- De-facto standard (FastAPI)
- Native type hints
- Better DX IDE/mypy


udata_hydra/schemas/check.py
```diff
@@ -1,42 +1,66 @@
...
+        if isinstance(value, dict):
+            return value
+        return json.loads(value) if value else None
+
+
+class CheckGroupBy(BaseModel):
+    # Allow creation from ORM objects (like asyncpg Record) by enabling from_attributes
+    model_config = ConfigDict(from_attributes=True)
+
+    value: str
```
ThibaudDauce, line 65: This cannot be just `str` since it's the response of the database with the correct column type

We could do this:

```suggestion
    value: int | float | bool | str | UUID | datetime | None
```

Or if we don't want to maintain the manual list we could coerce to string always with `str | None`, something like:

```
70 +    @field_validator("value", mode="before")
71 +    @classmethod
72 +    def transform_value(cls, value: object) -> str | None:
73 +        return None if value is None else str(value)
```

I pushed tests in 004d035285470e52279d0bbdf9ef055cd73c29f7 to reproduce the bug

udata_hydra/schemas/resource.py
```diff
@@ -1,28 +1,37 @@
...
+    last_modified: datetime.datetime
+    extras: dict | None = None
+    harvest: dict | None = None
+
+
+class ResourceSchema(BaseModel):
+    # Allow creation from ORM objects (like asyncpg Record) by enabling from_attributes
+    model_config = ConfigDict(from_attributes=True)
+
+    dataset_id: str
```
ThibaudDauce, line 33: `dataset_id` is nullable in database. Did you check all the values to ensure there is no null? And if it's the case maybe add a `NOT NULL` inside the database to be sure. Because if there is a null we'll have a 500 when reading this line.

udata_hydra/schemas/resource.py
```diff
@@ -1,28 +1,37 @@
...
-    dataset_id = fields.Str(required=True)
-    resource_id = fields.Str(required=True)
-    status = fields.Str(required=False)
-    status_since = fields.DateTime(required=False)
-    document = fields.Nested(ResourceDocumentSchema(), allow_none=True)
+    id: str
+    url: str
+    format: str | None = None
+    title: str
+    schema_name: str | None = None
```
ThibaudDauce, line 15: It was a `dict` from udata? Will it work?

Maybe use an alias if it's just to avoid "Field name "schema" in "WithSchema" shadows an attribute in parent "BaseModel""
```suggestion
    schema_: dict | None = Field(default=None, alias="schema")
```

udata_hydra/schemas/resource.py
```diff
@@ -1,28 +1,37 @@
...
+    filesize: int | None = None
+    checksum_type: str | None = None
+    checksum_value: str | None = None
+    created_at: datetime.datetime
+    last_modified: datetime.datetime
+    extras: dict | None = None
+    harvest: dict | None = None
+
+
+class ResourceSchema(BaseModel):
```
ThibaudDauce, line 29: You use the same document for input and output but they have really different attributes, no?

For exemple you need to manually check for "raise web.HTTPBadRequest(text="Missing document body")" because `document` is nullable because it's not present in the response (but require in the input?), and `status` is here but it cannot be accepted as input?

tests/test_api/test_api_checks.py
```diff
@@ -64,6 +64,11 @@ async def test_get_latest_check(setup_catalog, client, query, expected, fake_che
         "pmtiles_size": None,
         "geojson_url": None,
         "geojson_size": None,
+        "detected_last_modified_at": None,
+        "analysis_error": None,
+        "checksum": None,
+        "filesize": None,
+        "mime_type": None,
```
ThibaudDauce, line 71: Could be nice to add real tests for getting these values?

udata_hydra/routes/resources.py
```diff
@@ -99,15 +96,15 @@ async def update_resource(request: web.Request) -> web.Response:
...
-        dataset_id=dataset_id,
+        dataset_id=payload.dataset_id,
         resource_id=resource_id,
-        url=document["url"],
-        type=document["type"],
-        format=document["format"],
-        title=document["title"],
+        url=document.url,
+        type=document.type,
+        format=document.format or "",
```
ThibaudDauce, line 103: `format` is nullable inside database maybe better to set `None` than `""`?

udata_hydra/routes/resources.py
```diff
@@ -36,30 +36,28 @@ async def create_resource(request: web.Request) -> web.Response:
...
-        dataset_id=dataset_id,
-        resource_id=resource_id,
-        url=document["url"],
-        type=document["type"],
-        format=document["format"],
+        dataset_id=payload.dataset_id,
+        resource_id=str(payload.resource_id),
+        url=document.url,
+        type=document.type,
+        format=document.format or "",
```
ThibaudDauce, line 55: `format` is nullable inside database maybe better to set `None` than `""`?

udata_hydra/schemas/check.py
```diff
@@ -1,42 +1,66 @@
...
+        if value is None:
+            return None
+        if isinstance(value, dict):
+            return value
+        return json.loads(value) if value else None
+
+
+class CheckGroupBy(BaseModel):
+    # Allow creation from ORM objects (like asyncpg Record) by enabling from_attributes
+    model_config = ConfigDict(from_attributes=True)
```
ThibaudDauce, line 63: Are you sure these lines are useful? Claude seems to say we only need it on `CheckSchema` for `populate_by_name` but not on other models

udata_hydra/routes/resources_exceptions.py
```diff
@@ -26,71 +34,83 @@ async def create_resource_exception(request: web.Request) -> web.Response:
...
         resource_id=resource_id,
-        table_indexes=table_indexes,
-        comment=comment,
+        table_indexes=payload.table_indexes,
+        comment=payload.comment,
     )
 
-    return web.json_response(ResourceExceptionSchema().dump(dict(resource_exception)))  # type: ignore[arg-type]
+    if not resource_exception:
+        raise web.HTTPNotFound(text="Resource exception not found")
```
ThibaudDauce, line 101: I've added a test for this fix in 654ed85fd2afb82cbed33789a97d71532b17bfc1

ThibaudDauce (changes requested):

udata_hydra/schemas/resource.py
bolinocroustibat: There is no NULL dataset_id in production base so I added the constraint in migrations.

udata_hydra/schemas/resource.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/665d300c5f62cce24eb69be9b26f659d74e8edf5

udata_hydra/schemas/check.py
bolinocroustibat: Fixed here: https://github.com/datagouv/hydra/pull/465/changes/102c2d698a0a957cdf233400d56251e7b6fea6cc

tests/test_api/test_api_checks.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/d9556e62c2a9698456415754e30f6d38ac352203

udata_hydra/routes/resources.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/b0fb3259aabb57ab173fb6cf928989022b1f5a91

udata_hydra/routes/resources.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/b0fb3259aabb57ab173fb6cf928989022b1f5a91

udata_hydra/schemas/check.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/3a651eec19cc8eb5de38b58f5389ecaaa99588c6

udata_hydra/schemas/resource.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/6a46f49c21fe4c182aace5ad90f45442093af35e

udata_hydra/routes/checks.py
```diff
@@ -53,26 +58,28 @@ async def get_checks_aggregate(request: web.Request) -> web.Response:
...
     try:
-        resource: Record | None = await Resource.get(resource_id)
-        url: str = resource["url"]  # type: ignore[index]
+        resource: Record | None = await Resource.get(str(payload.resource_id))
+        if not resource:
+            raise web.HTTPNotFound(text=f"Couldn't find URL for resource {payload.resource_id}")
+        url: str = resource["url"]
     except Exception:
-        raise web.HTTPNotFound(text=f"Couldn't find URL for resource {resource_id}")
+        raise web.HTTPNotFound(text=f"Couldn't find URL for resource {payload.resource_id}")
```
ThibaudDauce: This guard catch every exception as a 404? It also catch and re-raise the exception `raise web.HTTPNotFound(text=f"Couldn't find URL for resource {payload.resource_id}")` just above?

I'm not sure to understand why this `except Exception` is useful?

udata_hydra/schemas/check.py
```diff
@@ -1,42 +1,69 @@
...
-
-class CheckGroupBy(Schema):
-    value = fields.Str()
-    count = fields.Integer()
+from pydantic import BaseModel, ConfigDict, Field, field_validator
+
+
+class CheckSchema(BaseModel):
+    # Allow population from field names (id, status) or aliases (check_id, check_status) by enabling populate_by_name
+    model_config = ConfigDict(populate_by_name=True)
```
ThibaudDauce: `populate_by_name` is deprecated (replaced by `validate_by_name`) but I'm not even sure why it's useful? `CheckSchema` is always validated from a `SELECT` with the correct names?

udata_hydra/routes/resources_exceptions.py
```diff
@@ -26,71 +34,83 @@ async def create_resource_exception(request: web.Request) -> web.Response:
...
+        request_data: dict = await request.json()
+        payload: UpdateResourceExceptionRequest = UpdateResourceExceptionRequest.model_validate(
+            request_data
+        )
+        if payload.table_indexes:
+            valid, error = ResourceExceptionSchema.are_table_indexes_valid(payload.table_indexes)
             if not valid:
                 raise web.HTTPBadRequest(text=error)
     except Exception as err:
         raise web.HTTPBadRequest(text=json.dumps({"error": str(err)}))
```
ThibaudDauce: Not really in this PR but the exception is raise and catch directly after, the client see
`{"error": "Bad Request"}` instead of `error, index type must be one of: index`.

ThibaudDauce (approved):
analysis/resource.py and analysis/helpers.py construct dict manually with `.isoformat()` so dates are not always the same 
- `2025-04-04T00:00:00Z` in `detected_last_modified_at`
- `2025-04-04T00:00:00+00:00` for the extra `analysis:last-modified-at` in `resource.py:237`

I think we should have the same datetime format everywhere (and maybe add this change in the PR description because it was not exactly the same format before this)

udata_hydra/routes/checks.py
bolinocroustibat: Good catch it's redundant indeed, simplified here: https://github.com/datagouv/hydra/pull/465/changes/66b2d9792300b3d5c067a9aa693e12a336daf991

bolinocroustibat:
> analysis/resource.py and analysis/helpers.py construct dict manually with `.isoformat()` so dates are not always the same
> 
> * `2025-04-04T00:00:00Z` in `detected_last_modified_at`
> * `2025-04-04T00:00:00+00:00` for the extra `analysis:last-modified-at` in `resource.py:237`
> 
> I think we should have the same datetime format everywhere (and maybe add this change in the PR description because it was not exactly the same format before this)

Good Point. To reduce the externalities this PR might induce, I suggest we adapt to the old format (`+00:00`) rather than Pydantic's one, so that it's exactly the same as seen from udata: https://github.com/datagouv/hydra/pull/465/changes/ba23c6d12f678646bd5b2840f1188024f91d90f9

We can simplify that later to use only `Z` in a later, separate PR.

udata_hydra/schemas/check.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/465/changes/fd9de94c74218e7ee85a8180d04498d5267f1110

udata_hydra/routes/resources_exceptions.py
bolinocroustibat: https://github.com/datagouv/hydra/pull/473

ThibaudDauce (approved):

PR #466, opened 2026-07-28
chore: add logs to trace check-notify gap on resource update

bolinocroustibat:
Add some debugs to help investigate https://github.com/datagouv/data.gouv.fr/issues/2048


maudetes (approved):
Thank you 🙏

Issue #468, opened 2026-08-05
Stop treating all `.gz` files as `csv.gz`


Pierlou:
Currently, if a file is gunzipped, there is no failsafe to detect whether it's a `csv.gz` or another format. This leads for instance to `json.gz` files being treated (sucessfully 🤯) as csv, but ends up displaying a clunky previz (eg [this resource](https://www.data.gouv.fr/datasets/finess-structures-1?resource_id=cd493959-fb03-41e5-9347-0edd14dfbc22)).
There should be a check (maybe just after the unzipping?) to assess the unzipped file's content type again, to prevent/perform further analysis accordingly.

bolinocroustibat:
This PR might solve this issue: https://github.com/datagouv/hydra/pull/469

bolinocroustibat:
Should be closed with https://github.com/datagouv/hydra/pull/476

PR #469, opened 2026-08-12
feat: analyse the single tabular file held in a zip archive

ThibaudDauce:
Allow analysing [Demandes de valeurs foncières](https://www.data.gouv.fr/datasets/demandes-de-valeurs-foncieres) and [Base Sirene des entreprises et de leurs établissements (SIREN, SIRET)](https://www.data.gouv.fr/datasets/base-sirene-des-entreprises-et-de-leurs-etablissements-siren-siret). Right now they will be blocked because of size but it's better than "Ce fichier ne peut pas être prévisualisé. Téléchargez-le depuis l'onglet Téléchargements."

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ Since it's called _hydra_, it also has mythical powers embedded:
 - if the remote resource is tabular (csv or excel-like), convert it to a PostgreSQL table, ready for APIfication, and to parquet to offer another distribution of the data
 - if the remote resource is a geojson, convert it to PMTiles to offer another distribution of the data
 - if the remote resource is parquet, ingest it into a PostgreSQL table for API exposition (similarly to csv/excel resources)
+- if the remote resource is a zip archive holding a single analysable file, extract it and analyse it as its own format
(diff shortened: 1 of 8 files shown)


bolinocroustibat:
Should close https://github.com/datagouv/hydra/issues/468

Pierlou:
FWIW: the SIREN files all have a parquet doppleganger (maintained by INSEE) which should be handled by hydra already (theoretically at least, RAM and/or timeouts may still be a problem) and the DVF files have [a cleaner (and geolocalized) csv.gz version](https://www.data.gouv.fr/datasets/demandes-de-valeurs-foncieres-geolocalisees?resource_id=d7933994-2c66-4131-a4da-cf7cd18040a4) and will have single file equivalent resources when [this](https://github.com/datagouv/datagouvfr_data_pipelines/tree/main/data_processing/dvf/geoloc) is ready, which should all be processed too.
Other usecases could be the zips from INSEE's Mélodi which contain two csv files (data + metadata), but that's rather niche

PR #470, opened 2026-08-12
perf: cast dates with the format inferred by csv-detective

ThibaudDauce:
Use the new `date_format` added in https://github.com/datagouv/csv-detective/pull/245

| Fichier                      | inspection |         |       | csv to db |         |           | total |         |           |
| ---------------------------- | ---------: | ------: | ----: | --------: | ------: | --------: | ----: | ------: | --------: |
|                              | main       | branche | Δ     | main      | branche | Δ         | main  | branche | Δ         |
| ValeursFoncieres-2024.txt    | 44,2       | 44,3    | +0 %  | 50,1      | 31,6    | **−37 %** | 94,3  | 75,9    | **−20 %** |
| ValeursFoncieres-2025-S1.txt | 16,4       | 16,5    | +0 %  | 20,2      | 12,7    | **−37 %** | 36,7  | 29,2    | **−20 %** |
| MN_07 (sans date)            | 30,5       | 28,9    | −5 %  | 9,9       | 10,1    | +2 %      | 40,3  | 39,0    | −3 %      |
| joconde.csv                  | 106,8      | 104,2   | −2 %  | 22,6      | 18,5    | **−18 %** | 129,4 | 122,8   | −5 %      |
| irve.csv                     | 7,9        | 6,7     | −15 % | 12,6      | 5,7     | **−55 %** | 20,6  | 12,4    | **−40 %** |

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ dependencies = [
     "aiohttp>=3.10.3",
     "asyncpg>=0.29.0",
     "coloredlogs>=15.0.1",
-    "csv-detective==0.12.0",
+    "csv-detective",
     "dateparser>=1.1.7",
     "humanfriendly>=10.0",
     "json-stream>=2.3.3",
@@ -93,6 +93,11 @@ local_scheme = "no-local-version"
 # Provisional: drop when sentry-sdk requires a patched urllib3 lower bound (GHSA-mf9v-mfxr-j63j).
 [tool.uv]
 constraint-dependencies = ["urllib3>=2.7.0"]
+
+# `cast()` only takes the inferred date format from that branch on. Pin it back to a released
+# version once datagouv/csv-detective#245 is merged and published.
+[tool.uv.sources]
+csv-detective = { git = "https://github.com/datagouv/csv-detective", branch = "improve_date_parsing" }
 
 [tool.uv.build-backend]
 module-name = "udata_hydra"
--- a/tests/test_conversion/test_csv_to_db.py
+++ b/tests/test_conversion/test_csv_to_db.py
@@ -81,6 +81,31 @@ async def test_csv_to_db_complex_type_casting(db, line_expected, clean_db, fake_
     assert dict(res[0]) == {k: v for k, v in zip(cols, expected)}
 
 
+@pytest.mark.parametrize(
+    "values_expected",
+    (
+        # a value no other value of the column disambiguates is read day-first
+        (["05/03/2022"], [date(2022, 3, 5)]),
+        # a single day-only value settles the format of the whole column...
+        (["03/04/2022", "25/04/2022"], [date(2022, 4, 3), date(2022, 4, 25)]),
+        # ...and the same value is read the other way around in a month-first column
+        (["03/04/2022", "04/25/2022"], [date(2022, 3, 4), date(2022, 4, 25)]),
+    ),
+)
+async def test_csv_to_db_ambiguous_date_column(db, values_expected, clean_db, fake_check):
+    check = await fake_check()
+    values, expected = values_expected
+    rows = "\n".join(f"{index};{value}" for index, value in enumerate(values, start=1))
+    with NamedTemporaryFile() as fp:
+        fp.write(f"int;date\n{rows}".encode("utf-8"))
+        fp.seek(0)
+        file = Csv(file_name=os.path.basename(fp.name), resource_id=RESOURCE_ID)
+        await file.inspect()
+        table = await file.to_db(check=check)
+    res = await db.fetch(f'SELECT date FROM "{table.table_name}" ORDER BY __id')
+    assert [row["date"] for row in res] == expected
+
+
 async def test_basic_sql_injection(db, clean_db, fake_check):
     check = await fake_check()
     # tries to execute
--- a/udata_hydra/utils/casting.py
+++ b/udata_hydra/utils/casting.py
@@ -9,14 +9,16 @@ from udata_hydra.utils.reader import Reader
 log = logging.getLogger("udata-hydra")
 
 
-def _smart_cast(_type: str, value, cast_json: bool = True, failsafe: bool = False) -> Any:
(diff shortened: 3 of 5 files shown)


pyproject.toml
```diff
@@ -12,7 +12,7 @@ dependencies = [
     "aiohttp>=3.10.3",
     "asyncpg>=0.29.0",
     "coloredlogs>=15.0.1",
-    "csv-detective==0.12.0",
+    "csv-detective",
```
bolinocroustibat, line 15: I would not unpin the csv detective version in pyproject, since it's a base component of hydra we need to track and pin the version explicitly by bumping it manually. The manual version change in the pyproject file is an established process in the repo and a good way to protect from human errors.

pyproject.toml
```diff
@@ -94,6 +94,11 @@ local_scheme = "no-local-version"
 [tool.uv]
 constraint-dependencies = ["urllib3>=2.7.0"]
 
+# `cast()` only takes the inferred date format from that branch on. Pin it back to a released
```
bolinocroustibat, line 97: I would prefer not to add that kind of shim to the version control which decreases readability, maintainability and not as explicit ad the version in pyproject.
Better to create a release of csv-detective and pin the version of the module in pyproject so that it's very explicit.

tests/test_conversion/test_csv_to_db.py
```diff
@@ -81,6 +81,31 @@ async def test_csv_to_db_complex_type_casting(db, line_expected, clean_db, fake_
     assert dict(res[0]) == {k: v for k, v in zip(cols, expected)}
 
 
+@pytest.mark.parametrize(
+    "values_expected",
+    (
+        # a value no other value of the column disambiguates is read day-first
```
bolinocroustibat, line 87: What about the date format YYYY/MM/DD? Is it supported by csv-detective? In any case, should we test it?

udata_hydra/utils/casting.py
```diff
@@ -29,27 +31,31 @@ def iter_tabular_rows(
...
                     yield [
                         _smart_cast(
                             _type,
+                            date_format,
                             value if isinstance(value, str) or value is None else str(value),
                             cast_json=cast_json,
                             failsafe=False,
                         )
-                        for _type, value in zip(columns.values(), line)
+                        for (_type, date_format), value in zip(columns.values(), line)
```
bolinocroustibat, line 49: This is becoming a bit difficult to read so I changed for a more explicit (and performant) unpacking: https://github.com/datagouv/hydra/pull/470/changes/bb5d99170c565e5e8f5588166015efcc5ea8805d

bolinocroustibat (changes requested):
Thanks! A few remarks, an important one is to keep the explicit csv-detective version in pyproject (and make a release of csv-detective if necessary).

pyproject.toml
ThibaudDauce: I think it can be reverted as soon as the new version is released (this hydra branch require the new csv-detective of https://github.com/datagouv/csv-detective/pull/245)

pyproject.toml
ThibaudDauce: I don't understand, this PR require the new version of csv-detective and cannot be merge without it… This comment is just here to remind us to change this, no?

pyproject.toml
bolinocroustibat: I would not merge this branch before there is a version of csv-detective released for that!

pyproject.toml
bolinocroustibat: I would not merge this branch before there is a version of csv-detective released for that!

tests/test_conversion/test_csv_to_db.py
estellebertrand: Yes this is supposed to be supported ! As well as YYYY/DD/MM

estellebertrand: Agree we should test it

tests/test_conversion/test_csv_to_db.py
```diff
@@ -81,6 +81,31 @@ async def test_csv_to_db_complex_type_casting(db, line_expected, clean_db, fake_
...
+        (["05/03/2022"], [date(2022, 3, 5)]),
+        # a single day-only value settles the format of the whole column...
+        (["03/04/2022", "25/04/2022"], [date(2022, 4, 3), date(2022, 4, 25)]),
+        # ...and the same value is read the other way around in a month-first column
+        (["03/04/2022", "04/25/2022"], [date(2022, 3, 4), date(2022, 4, 25)]),
```
estellebertrand:
```suggestion
        (["03/04/2022", "04/25/2022"], [date(2022, 3, 4), date(2022, 4, 25)]),
        # same for date starting by year instead
        (["2022/03/04", "2022/25/04"], [date(2022, 4, 3), date(2022, 4, 25)]),
        (["2022/03/04", "2022/04/25"], [date(2022, 3, 4), date(2022, 4, 25)]),
```

like that?

tests/test_conversion/test_csv_to_db.py
ThibaudDauce: Added in 3393ef2 with a year-first column where the ambiguous value is settled by another one (`2022/03/05` + `2022/25/04` is read `%Y/%d/%m` so the first one is May 3rd). It fails without the inferred format, a plain `YYYY/MM/DD` column was passing with and without the change.

I didn't add the other formats, csv-detective tests them already and here we only need to check the format is passed to `cast()`?

ThibaudDauce:
@bolinocroustibat @estellebertrand is there a way to publish the branch version on preproduction to test it on real data? We can deploy the artifact https://output.circle-artifacts.com/output/job/d7f7ddc4-315b-435e-8461-00d48e35ea5e/artifacts/0/dist/udata_hydra-2.13.6.dev13-py3-none-any.whl directly with udata but don't know if it works the same in hydra

We could let it run a few days to see if there is any regression?

